PR #5, opened 2021-09-30
improve handling of poweron/poweroff for LPARs and VIOSes

dberg1:


app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm.rb
```diff
@@ -1,49 +1,36 @@
...
+    power
   end
 
   def raw_stop
     $ibm_power_hmc_log.info("raw_stop ems_ref=#{ems_ref}")
-    ext_management_system.with_provider_connection do |connection|
-      # Damien: check VIOS or LPAR from description?
-      connection.poweroff_lpar(ems_ref, {"operation" => "shutdown"})
-    end
+    power({"operation" => "shutdown"})
```
Fryguy, line 18: I don't know why the bot is not flagging this, but the curly braces are redundant

```suggestion
    power("operation" => "shutdown")
```

Same with the other calls below.

app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm.rb
```diff
@@ -76,4 +63,28 @@ def raw_rename
...
+          connection.poweroff_lpar(ems_ref, params)
+        end
+      elsif vios?
+        connection.poweron_vios(ems_ref, params)
+      else
+        connection.poweron_lpar(ems_ref, params)
+      end
+    end
+  rescue IbmPowerHmc::Connection::HttpError => e
+    $ibm_power_hmc_log.error("error changing power state of #{ems_ref} with params=#{params}: #{e}")
```
Fryguy, line 88: Should this re-raise the exception?  As is, this causes the method, on failure, to return `true`

app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm.rb
```diff
@@ -76,4 +63,28 @@ def raw_rename
...
     POWER_STATES[raw_power_state] || "unknown"
   end
+
+  private
+
+  def vios?
+    description == "Virtual IO Server"
+  end
+
+  def power(params = {})
```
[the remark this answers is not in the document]
agrare:
```ruby
def poweroff(params = {})
  ext_management_system.with_provider_connection do |connection|
    connection.send("poweroff_#{vios? ? "vios" : "lpar"}", params)
  end
end

def poweron
  ext_management_system.with_provider_connection do |connection|
    connection.send("poweron_#{vios? ? "vios" : "lpar"}"
  end
end
```

PR #7, opened 2021-10-01
Ensure HMC connection logoff

vfebvre:


app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_catcher/stream.rb
```diff
@@ -31,5 +31,7 @@ def poll
         raise ProviderUnreachable, exception.message
       end
     end
+  ensure
+    connection.logoff
```
dberg1, line 35: For all places that use `with_provider_connection`, you should just override the `ExtManagementSystem.disconnect` method in in `manageiq-providers-ibm_power_hmc/app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb` and call `connection.logoff`. The `diconnect` method will be called automatically by `with_provider_connection`.
Something like:

```
def disconnect(connection)
  connection.logoff
end
```
For information, this is what `with_provider_connection` does:

```
  def with_provider_connection(options = {})
    raise _("no block given") unless block_given?
    _log.info("Connecting through #{self.class.name}: [#{name}]")
    connection = connect(options)
    yield connection
  ensure
    disconnect(connection) if connection
  end
```

dberg1:
LGTM

PR #11, opened 2021-10-14
add partition id and reference code as read-only vm advanced settings

dberg1:
Partition id and reference code are PowerVM specific and do not really fit in the MIQ database schema.
Add them as read-only advanced settings.
They are quite useful though and are displayed by the HMC UI like so:
image


dberg1:
Need to investigate if the reference code better fits in the connection_state column.
Examples of what reference codes can be:
https://www.ibm.com/docs/en/power5?topic=POWER5/ipha6_p5/x20a8atx.htm

dberg1:
Looking at app/models/vm_or_template.rb, it would seem that if `connection_state` is set to anything other than `"connected"`, the VM would be considered as disconnected, which is not what we want here.

vfebvre:
@dberg1 HMC SDK gem v0.2.1 pushed

PR #16, opened 2021-10-15
[WIP] retrieve cpu_speed for hosts

dberg1:
Not sure if we want this.
This seems quite a bit silly to go to these lengths just to get the cpu frequency, but it does not appear to be anywhere in the XML documents returned by the HMC.
I believe you can get it through PCM REST API but PCM is not enabled by default on most managed systems and comes at a cost.
I don't see it reported in the HMC UI either.
Wondering where PowerVC gets it from, but it is displayed in the PowerVC UI:

image

In MIQ, with these changes, it shows as:

image


dberg1:
Closing as this is probably not a good idea.

PR #23, opened 2021-11-26
Add Host Ethernet Adapter Logical Ports to inventory

vfebvre:
Another type of ethernet adapters.
These do not have UUIDs so we use the MAC address as the unique ID instead.


jaywcarman (approved):
LGTM

Issue #8, opened 2021-10-07
[Najdorf] IBM Power HMC Enhancements


agrare:
- [x] Expand inventory collection
  - [x] Virtual Switches - https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/pull/17
  - [x] Networks - https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/pull/17
  - [x] Ethernet Adapters - https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/pull/18
  - [x] Templates - https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/pull/22
  - [x] Logical Ports - https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/pull/23
  - [x] Storages - https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/pull/31
- [x] Event handler - https://github.com/ManageIQ/manageiq-content/pull/692
- [x] Provision
  - [x] Vm to Template https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/pull/26

jaywcarman:
Deferring metrics collection to a subsequent release.

PR #26, opened 2021-12-09
Add template provisioning

vfebvre:


app/models/manageiq/providers/ibm_power_hmc/infra_manager/provision.rb
```diff
@@ -0,0 +1,7 @@
+class ManageIQ::Providers::IbmPowerHmc::InfraManager::Provision < MiqProvision
+  include_concern 'Cloning'
+  include_concern 'StateMachine'
+  def destination_type
+    "Vm"
```
agrare, line 5: If you're provisioning _to_ a template then the destination_type should be `"Template"

agrare: You should also validate that the request_type is `"clone_to_template"`, see https://github.com/ManageIQ/manageiq-providers-vmware/blob/master/app/models/manageiq/providers/vmware/infra_manager/provision.rb#L9 for examples

chessbyte:
@jaywcarman @agrare should this be part of an epic on the ManageIQ roadmap?  also, is there anything to update in the [Capabilities Matrix](https://www.manageiq.org/docs/reference/latest/capabilities_matrix)?

agrare:
@chessbyte is it part of the https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/issues/8 enhancement tracking issue which is on the roadmap.  An update to the capabilities matrix would be good yes

PR #24, opened 2021-12-06
[WIP] SharedStoragePool inventory

CallisteH:
SSPs are inserted inside `storages` table in the MIQ model.

For now, only the global refresh works. This means that _ssps_ are inserted at the MIQ startup.
image

PR #31, opened 2022-01-26
SharedStoragePool inventory

vfebvre:
Recreate new PR for https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/pull/24

### DIFF
--- a/app/models/manageiq/providers/ibm_power_hmc/inventory/collector/infra_manager.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/inventory/collector/infra_manager.rb
@@ -62,9 +62,7 @@ class ManageIQ::Providers::IbmPowerHmc::Inventory::Collector::InfraManager < Man
   attr_accessor :ssps
 
   def ssp_lus_by_udid
-    @ssp_lus_by_udid ||= {}
-    ssps.each { |ssp| ssp.lus.each { |lu| @ssp_lus_by_udid[lu.udid] = ssp.cluster_uuid } }
-    @ssp_lus_by_udid
+    @ssp_lus_by_udid ||= ssps.flat_map { |ssp| ssp.lus.map { |lu| [lu.udid, ssp.cluster_uuid] } }.to_h
   end
 
   def vscsi_lun_mappings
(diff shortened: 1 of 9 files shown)


agrare:
@vfebvre looks like you have to define a `Storage` subclass under your infra_manager

app/models/manageiq/providers/ibm_power_hmc/inventory/collector/infra_manager.rb
```diff
@@ -57,8 +59,30 @@ def templates
     @templates || []
   end
 
+  attr_accessor :ssps
+
+  def ssp_lus_by_udid
+    @ssp_lus_by_udid ||= {}
+    ssps.each { |ssp| ssp.lus.each { |lu| @ssp_lus_by_udid[lu.udid] = ssp.cluster_uuid } }
+    @ssp_lus_by_udid
```
agrare, line 67: So the way this is written it will redefine the `@ssp_lus_by_udid` everytime this is called.

You probably want something like:
```
@ssp_lus_by_udid ||= ssps.each_with_object({}) { |ssp, result| ssp.lus.each { |lu| result[lu.udid] = ssp.cluster_uuid } }
```

There's probably a cleaner way to one-line that but it is escaping me at the moment

chessbyte: There is also the begin/end way of doing this (if a one-liner does not work for some reason):

```ruby
@ssp_lus_by_udid ||= begin
  hash = {}
  ssps.each { |ssp| ssp.lus.each { |lu| hash[lu.udid] = ssp.cluster_uuid } }
  hash
end
```

Having said that, I think the one-liner is clearer and more appropriate in this case.

agrare: Something like `@ssp_lus_by_udid ||= ssps.flat_map { |ssp| ssp.lus.map { |lu| [lu.udid, ssp.cluster_uuid] } }.to_h` should do the trick also

agrare (approved):
:+1: LGTM

PR #32, opened 2022-01-26
Fix Templates not subclassed under IbmPowerHmc

agrare:
These were being created as `ManageIQ::Providers::InfraManager::Template` types

### DIFF
--- a/app/models/manageiq/providers/ibm_power_hmc/inventory/collector/target_collection.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/inventory/collector/target_collection.rb
@@ -115,7 +115,7 @@ class ManageIQ::Providers::IbmPowerHmc::Inventory::Collector::TargetCollection <
         add_target!(:host_virtual_switches, target.ems_ref)
       when Lan
         add_target!(:lans, target.ems_ref)
-      when ManageIQ::Providers::InfraManager::Template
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/ibm_power_hmc/inventory/collector/target_collection.rb
```diff
@@ -115,7 +115,7 @@ def parse_targets!
         add_target!(:host_virtual_switches, target.ems_ref)
       when Lan
         add_target!(:lans, target.ems_ref)
-      when ManageIQ::Providers::InfraManager::Template
+      when ManageIQ::Providers::IbmPowerHmc::InfraManager::Template
```
jaywcarman, line 118: Why do we use the full namespace for some of these and the base class for others? Seems to be the same for the IBM Cloud providers as well.

agrare: Yeah we actually don't need to, you can use the base class in a `when` condition and children will match it, I was just staying consistent with the others (except Host).

jaywcarman: Makes sense, thanks for helping me to understand!

jaywcarman (approved):
LGTM

PR #38, opened 2022-02-11
Fix vendor references

jaywcarman:
Use "ibm_power_hmc" vendor name to match the existing decorator icon.

Depends on: https://github.com/ManageIQ/manageiq/pull/21701


Fryguy:
Skipping backport to `najdorf`, because it is already in the branch.

PR #45, opened 2022-03-02
Fix verify_credentials for an existing EMS

agrare:
When an EMS is being edited the password is not sent back down unless it is changed so we have to look it up when verifying credentials.

This was failing because of an undefined variable `authtype`.
2696c280-995f-11ec-974d-0d9e8bf2dd19

### DIFF
--- a/app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
@@ -106,7 +106,7 @@ class ManageIQ::Providers::IbmPowerHmc::InfraManager < ManageIQ::Providers::Infr
     authentication = args.dig("authentications", "default")
     userid, password = authentication&.values_at("userid", "password")
     password = ManageIQ::Password.try_decrypt(password)
-    password ||= find(args["id"]).authentication_password(authtype) if args['id']
+    password ||= find(args["id"]).authentication_password("default") if args['id']
 
     !!raw_connect(hostname, port, userid, password, validate_ssl, true)
   end


app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
```diff
@@ -106,7 +106,7 @@ def self.verify_credentials(args)
     authentication = args.dig("authentications", "default")
     userid, password = authentication&.values_at("userid", "password")
     password = ManageIQ::Password.try_decrypt(password)
-    password ||= find(args["id"]).authentication_password(authtype) if args['id']
+    password ||= find(args["id"]).authentication_password("default") if args['id']
```
agrare, line 109: NOTE we would use a dynamic `authtype` if more types of authentications were required (e.g. multiple endpoints with different services and credentials) but [[up here]](https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/pull/45/files#diff-7c3c4c9da3d6d2f60b9c2853dd82040d585e3541aea2640df4b900dca8c29acaR106) we're only using `"default"`.

jaywcarman (approved):
Thanks @agrare !

Fryguy:
Backported to `najdorf` in commit d49f2f070dac27930d850c243963b1aa563f80bc.

```text
commit d49f2f070dac27930d850c243963b1aa563f80bc
Author: Jay Carman <jwcarman@us.ibm.com>
Date:   Wed Mar 2 07:38:00 2022 -0600

    Merge pull request #45 from agrare/fix_verify_credentials_edit
    
    Fix verify_credentials for an existing EMS
    
    (cherry picked from commit dab4ae74305aab365e97864d3ddb1b2b9e17f749)
```

PR #46, opened 2022-03-08
updated version

### LINKED ISSUE #44: Improve error message when provider connection fails
When a connection to the HMC endpoint fails, the raw exception from the `ibm_power_hmc_sdk_ruby` module is raised and the user ends up with a confusion message. This happened during test when an HMC "went to lunch" and returned http 408 timeout. The following messages were displayed in the GUI:
```
Validation failed: Malformed DOCTYPE: garbage after public ID literal Line: 7 Position: 221 Last 80 unconsumed characters: PUBLIC "-//IETF//DTD HTML 2.0//EN"> <html><head> <title>408 Request Timeout</tit
```
and
```
Validation failed: Declarations can only occur in the doctype declaration. Line: 1 Position: 333 Last 80 unconsumed characters: <!doctype html public "-//IETF//DTD HTML 2.0//EN"> <html><head><title>Console Int
```

I think it would be helpful to add a `rescue` to the `raw_connect` method (https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/blob/master/app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb#L140-L151) and construct a message like "HMC connection failed: <add some parsed details>" before re-raising the exception.

For reference, here's something similar from PowerVS:
https://github.com/ManageIQ/manageiq-providers-ibm_cloud/blob/master/app/models/manageiq/providers/ibm_cloud/power_virtual_servers/manager_mixin.rb#L137-L165

CallisteH:
fixes #44


CallisteH:
This error from the `ci` can be fixed locally if I run `bin/setup` before. But not on this pull-request I guess ? 

Following error : 
```
$ bundle exec rake
rake aborted!
FrozenError: can't modify frozen Hash: {"__bootsnap_ruby_version__"=>"5445e0435260b449decf2ac16f9d09bae3cafe72-powerpc64le-linux", "/home/padmin/Calliste/manageiq-providers-ibm_power_hmc/.bundle/plugin/gems/bundler-inject-2.0.0/lib"=>[1644827107, ["bundler-inject.rb", "bundler/inject.rb", "bundler/inject/dsl_patch.rb", "bundler/inject/version.rb"], ["bundler", "bundler/inject"]], "/home/padmin/Calliste/manageiq-providers-ibm_power_hmc/lib"=>[1644239464, ["manageiq-providers-ibm_power_hmc.rb", "manageiq/providers/ibm_power_hmc/engine.rb", "manageiq/providers/ibm_power_hmc/version.rb"], ["manageiq", "manageiq/providers", "manageiq/providers/ibm_power_hmc", "tasks", "tasks_private"]]

[...]
```

agrare:
@CallisteH that looks like a sporadic failure, please drop the `test` commit and repush

PR #58, opened 2022-05-05
Feature/power operations

CallisteH:
We add the ability to turn off or turn on a host.


CallisteH:
### 1
image
### 2
image

Logs : 
`ERROR -- evm: MIQ(ems_infra_controller-button): "Power Off": not available for pollux`

app/models/manageiq/providers/ibm_power_hmc/infra_manager/host.rb
```diff
@@ -1,6 +1,8 @@
 class ManageIQ::Providers::IbmPowerHmc::InfraManager::Host < ::Host
   include ManageIQ::Providers::IbmPowerHmc::InfraManager::MetricsCaptureMixin
 
+  supports :power_off
```
CallisteH, line 4: @agrare I don't know why I have this error in the logs but I wonder if it's not because I'm using the wrong keyword here?

agrare: Hey @CallisteH we have to have `supports_not :feature_name` in the base class (in this case `::Host`) so that we can ask `instance.supports?(:feature_name)` on an instance of any class

CallisteH:
Working better with the last commit : 
image

app/models/manageiq/providers/ibm_power_hmc/infra_manager/host.rb
```diff
@@ -1,6 +1,42 @@
...
+      raise
+    end
+
+  end
+
+  def start
+    $ibm_power_hmc_log.info("#{self.class}##{__method__}")
+    ext_management_system.with_provider_connection do |connection|
+      connection.poweron_managed_system(ems_ref,
+        :params   => {"operation" => on}
```
agrare, line 22: Really minor but if you're not aligning the hash to anything best to leave out the extra padding
```suggestion
        :params => {"operation" => on}
```

app/models/manageiq/providers/ibm_power_hmc/inventory/parser/infra_manager.rb
```diff
@@ -20,7 +20,8 @@ def parse_cecs
         :hostname            => sys.hostname,
         :ipaddress           => sys.ipaddr,
         :power_state         => lookup_power_state(sys.state),
-        :vmm_vendor          => "ibm_power_hmc"
+        :vmm_vendor          => "ibm_power_hmc",
+        :type                => ManageIQ::Providers::IbmPowerHmc::InfraManager::Host.name
```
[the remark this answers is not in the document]
agrare: Just ran the specs:
```
ems.hosts.first
#<ManageIQ::Providers::IbmPowerHmc::InfraManager::Host
```

app/models/manageiq/providers/ibm_power_hmc/infra_manager/host.rb
```diff
@@ -1,6 +1,42 @@
 class ManageIQ::Providers::IbmPowerHmc::InfraManager::Host < ::Host
   include ManageIQ::Providers::IbmPowerHmc::InfraManager::MetricsCaptureMixin
 
+  supports :stop
+  supports :shutdown
```
agrare, line 5: Hey @CallisteH it is turning out to be quite the ordeal to convert Host to supports? from validate_*

For the time being I think it will be faster for us to do
```suggestion
  def validate_stop
    {:available => true,  :message => nil}
  end
  
  def validate_shutdown
    {:available => true,  :message => nil}
  end
```

Until we are able to move everything over to supports

app/models/manageiq/providers/ibm_power_hmc/infra_manager/host.rb
```diff
@@ -1,6 +1,42 @@
...
+
+  def shutdown
+    $ibm_power_hmc_log.info("#{self.class}##{__method__}")
+    ext_management_system.with_provider_connection do |connection|
+      connection.poweroff_managed_system(ems_ref)
+    rescue IbmPowerHmc::Connection::HttpError => e
+      $ibm_power_hmc_log.error("error powering off LPAR #{ems_ref} with params=#{params}: #{e}")
```
dberg1, line 12: params is not defined here.
Also it is not an LPAR so the text should be:
```suggestion
      $ibm_power_hmc_log.error("error powering off managed system #{ems_ref}: #{e}")
```

app/models/manageiq/providers/ibm_power_hmc/infra_manager/host.rb
```diff
@@ -1,6 +1,42 @@
...
+  end
+
+  def start
+    $ibm_power_hmc_log.info("#{self.class}##{__method__}")
+    ext_management_system.with_provider_connection do |connection|
+      connection.poweron_managed_system(ems_ref,
+        :params   => {"operation" => on}
+      )
+    rescue IbmPowerHmc::Connection::HttpError => e
+      $ibm_power_hmc_log.error("error powering off LPAR #{ems_ref} with params=#{params}: #{e}")
```
dberg1, line 25: same as above (params and LPAR):
```suggestion
      $ibm_power_hmc_log.error("error powering off managed system #{ems_ref}: #{e}")
```

app/models/manageiq/providers/ibm_power_hmc/infra_manager/host.rb
```diff
@@ -1,6 +1,42 @@
...
+  end
+
+  def stop
+    $ibm_power_hmc_log.info("#{self.class}##{__method__}")
+    ext_management_system.with_provider_connection do |connection|
+      connection.poweroff_managed_system(ems_ref,
+        :params   => {"immediate" => true}
+      )
+    rescue IbmPowerHmc::Connection::HttpError => e
+      $ibm_power_hmc_log.error("error powering off LPAR #{ems_ref} with params=#{params}: #{e}")
```
dberg1, line 36: same as above (params and LPAR):
```suggestion
      $ibm_power_hmc_log.error("error powering off managed system #{ems_ref}: #{e}")
```

app/models/manageiq/providers/ibm_power_hmc/infra_manager/host.rb
```diff
@@ -1,6 +1,42 @@
...
+    end
+  end
+
+  def stop
+    $ibm_power_hmc_log.info("#{self.class}##{__method__}")
+    ext_management_system.with_provider_connection do |connection|
+      connection.poweroff_managed_system(ems_ref,
+        :params   => {"immediate" => true}
+      )
```
dberg1, line 34: This can be simplified like so:
```suggestion
      connection.poweroff_managed_system(ems_ref, "immediate" => "true")
```

app/models/manageiq/providers/ibm_power_hmc/infra_manager/host.rb
```diff
@@ -1,6 +1,42 @@
...
+    ext_management_system.with_provider_connection do |connection|
+      connection.poweroff_managed_system(ems_ref)
+    rescue IbmPowerHmc::Connection::HttpError => e
+      $ibm_power_hmc_log.error("error powering off LPAR #{ems_ref} with params=#{params}: #{e}")
+      raise
+    end
+
+  end
+
+  def start
```
dberg1, line 18: I guess you need a `validate_start` too.

dberg1 (changes requested):

CallisteH:
@dberg1 Maybe we can develop a method that returns "true" if all lpars are off?

app/models/manageiq/providers/ibm_power_hmc/infra_manager/host.rb
```diff
@@ -1,6 +1,46 @@
...
+    {:available => true, :message => nil}
+  end
+
+  def validate_shutdown
+    {:available => true, :message => nil}
+  end
+
+  def validate_start
+    {:available => true, :message => nil}
+  end
```
agrare, line 14: @CallisteH you probably want to check the current power state of the host (e.g. you can't start a host that is already on)

CallisteH: Hi Adam, is this a good way to do it like this?
For example : 
```
  def validate_shutdown
    {:available => self.vms_off, :message => nil}
  end
```

where `vms_off` is the boolean function that returns true if all vms are shutdown

dberg1: @CallisteH You'd want to set a specific message too (like "Cannot shutdown Host while some VMs are still running").

dberg1: @CallisteH For your vm_offs method, you cannot perform REST requests to the HMC here! You need to look at the VM states in the inventory instead.

agrare:
Looking good @CallisteH you probably don't need VCR for the specs, you can `let(:conn) { double("PowerHMC") }` and then  `allow(ems).to receive(:with_provider_connection).and_yield(conn)` and then `expect(conn).to receive(:poweron_server).with(physical_server.ems_ref)`

We should check the `validates` methods in tests also by setting the power_state to different values confirming it works as expected.

CallisteH:
Hi @agrare 

Thanks for your review ! 

There are two or three elements that I don't quite understand.

```
  context "power_operation" do
    let(:conn) { double("IbmPowerHmc") } do
      it "start" do
        allow(ems).to receive(:with_provider_connection).and_yield(conn)
        expect(conn).to receive(:poweron_managed_system).with(host.ems_ref)
      end
    end
  end
```

How can I control what `poweron_managed_system` returned ? Where can I call `host.start` function ?

CallisteH:
@agrare, Thank you for your help. I made the changes accordingly

dberg1:
@CallisteH I see that https://github.com/ManageIQ/manageiq-ui-classic/pull/8266 has been merged.
I believe this means you have to use `supports` instead of implementing the `validate_xyz` methods.

app/models/manageiq/providers/ibm_power_hmc/infra_manager/host.rb
```diff
@@ -1,6 +1,47 @@
 class ManageIQ::Providers::IbmPowerHmc::InfraManager::Host < ::Host
   include ManageIQ::Providers::IbmPowerHmc::InfraManager::MetricsCaptureMixin
 
+  supports :stop do
+    unsupported_reason_add(:stop, _("Cannot shutdown a host that is powered off")) unless power_state="on"
```
agrare, line 5:
```suggestion
    unsupported_reason_add(:stop, _("Cannot shutdown a host that is powered off")) unless power_state == "on"
```

agrare:
Looks pretty good @CallisteH can you fix the conflict?

CallisteH:
@agrare I removed the [WIP] label in the title

app/models/manageiq/providers/ibm_power_hmc/inventory/parser/infra_manager.rb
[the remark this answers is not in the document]
agrare: Still think this isn't required but we can test in some specs in a follow-up

agrare (approved):

agrare:
@CallisteH one more thing before I merge, can you rebase and edit these commits to remove the WIP from the commit message?  No need to squash them just rewrite the message.

CallisteH:
@agrare thanks for your suggestion, everything looks good ?

PR #60, opened 2022-05-19
Check if PCM is enabled on host

vfebvre:
Retrieve Performance and Capacity Monitor preferences for each Host and store it so ManageIQ does not try to collect metrics for a host and its VMs if there is nothing to collect.

### DIFF
--- a/app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
@@ -19,6 +19,8 @@ class ManageIQ::Providers::IbmPowerHmc::InfraManager < ManageIQ::Providers::Infr
   supports :metrics
   supports :provisioning
 
+  has_many :hosts_advanced_settings, :through => :hosts, :source => :advanced_settings
+
   def self.params_for_create
     @params_for_create ||= {
       :fields => [
--- a/app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
@@ -65,25 +65,39 @@ class ManageIQ::Providers::IbmPowerHmc::InfraManager::EventTargetParser
   end
 
   def handle_usertask(usertask)
-    r = []
     if usertask["status"].eql?("Completed")
       case usertask["key"]
       when "TEMPLATE_PARTITION_SAVE", "TEMPLATE_PARTITION_SAVE_AS", "TEMPLATE_PARTITION_CAPTURE"
-        r << {:assoc => :miq_templates, :ems_ref => usertask['template_uuid']}
+        handle_usertask_template_save(usertask)
       when "TEMPLATE_DELETE"
-        template = ManageIQ::Providers::InfraManager::Template.find_by(:ems_id => ems_event.ext_management_system.id, :name => usertask['labelParams'])
-        unless template.nil?
-          r << {:assoc => :miq_templates, :ems_ref => template.uid_ems}
-        end
+        handle_usertask_template_delete(usertask)
       when "PCM_PREFERENCE_UPDATE"
-        usertask['labelParams'].first.tr("[] ", "").split(",").each do |hostname|
-          host = ManageIQ::Providers::IbmPowerHmc::InfraManager::Host.find_by(:ems_id => ems_event.ext_management_system.id, :name => hostname)
-          unless host.nil?
-            r << {:assoc => :hosts, :ems_ref => host.ems_ref}
-          end
-        end
+        handle_usertask_pcm_preference(usertask)
+      else
+        []
       end
     end
-    r
+  end
+
+  def handle_usertask_template_save(usertask)
+    [{:assoc => :miq_templates, :ems_ref => usertask['template_uuid']}]
+  end
+
+  def handle_usertask_template_delete(usertask)
+    template = ManageIQ::Providers::InfraManager::Template.find_by(:ems_id => ems_event.ext_management_system.id, :name => usertask['labelParams'])
+    if template.nil?
+      []
+    else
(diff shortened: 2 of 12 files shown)


app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
```diff
@@ -19,6 +19,8 @@ class ManageIQ::Providers::IbmPowerHmc::InfraManager < ManageIQ::Providers::Infr
   supports :metrics
   supports :provisioning
 
+  has_many :hosts_advanced_settings, :through => :hosts, :source => :advanced_settings
```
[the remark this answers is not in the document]
agrare: The base `Host` model already has advanced settings, but in order to pull these with refresh we need this association on the top-level manager.

For example here is the equivalent for VMs/Templates: https://github.com/ManageIQ/manageiq/blob/master/app/models/ext_management_system.rb#L122

I'm good with this here but eventually we could move this up to the base `ExtManagementSystem` model and add the inventory_collection to the core set.

app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
```diff
@@ -70,14 +68,36 @@ def handle_usertask(usertask)
...
+    if template.nil?
+      []
+    else
+      [{:assoc => :miq_templates, :ems_ref => template.uid_ems}]
+    end
+  end
+
+  def handle_usertask_pcm_preference(usertask)
+    usertask['labelParams'].first.tr("[] ", "").split(",").each_with_object([]) do |hostname, array|
+      host = ManageIQ::Providers::IbmPowerHmc::InfraManager::Host.find_by(:ems_id => ems_event.ext_management_system.id, :name => hostname)
```
agrare, line 97: Minor but it is better to use the association rather than the direct IDs (if the assoc is cached it can use that and not do a query)

```suggestion
      host = ManageIQ::Providers::IbmPowerHmc::InfraManager::Host.find_by(:ext_management_system => ems_event.ext_management_system, :name => hostname)
```

app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
```diff
@@ -70,14 +68,36 @@ def handle_usertask(usertask)
...
+      unless host.nil?
+        array << {:assoc => :hosts, :ems_ref => host.ems_ref}
```
agrare, line 99: I'd probably write this a little differently to reduce the number of queries:
```suggestion
    hostnames     = usertask['labelParams'].first.tr("[] ", "").split(",")
    host_ems_refs = ManageIQ::Providers::IbmPowerHmc::InfraManager::Host.where(:ems_id => ems_event.ext_management_system.id, :name => hostnames).pluck(:ems_ref)
    host_ems_refs.map { |ems_ref| {:assoc => :hosts, :ems_ref => ems_ref}
```

app/models/manageiq/providers/ibm_power_hmc/infra_manager/host.rb
```diff
@@ -1,6 +1,10 @@
 class ManageIQ::Providers::IbmPowerHmc::InfraManager::Host < ::Host
   include ManageIQ::Providers::IbmPowerHmc::InfraManager::MetricsCaptureMixin
 
+  supports :capture do
+    unsupported_reason_add(:capture, _("PCM not enabled for this Host")) unless advanced_settings.find { |s| s.name.eql?("pcm_enabled") }&.value.eql?("true")
```
agrare, line 5: I think it'd be helpful to have this as a virtual_column in the host model...try this out:
```
def pcm_enabled
  !!advanced_settings.detect { |s| s.name == "pcm_enabled" }&.value
end
virtual_column :pcm_enabled, :type => :boolean, :uses => :advanced_settings
```

app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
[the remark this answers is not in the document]
vfebvre: Yes I meant the Ems base class

app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
vfebvre: ok

app/models/manageiq/providers/ibm_power_hmc/infra_manager/host.rb
vfebvre: Fixed it, I think

spec/factories/advanced_setting.rb
```diff
@@ -0,0 +1,4 @@
+FactoryBot.define do
+  factory :ibm_power_hmc_advanced_setting,
+          :class => "AdvancedSetting"
+end
```
agrare, line 4: @vfebvre it is a little confusing that this is called `ibm_power_hmc_advanced_setting` when it is actually just the base class.  Since you aren't setting anything in the factory FactoryBot can "figure it out" from the name and define the factory on the fly.  Try just deleting this file, and using just `FactoryBot.create(:advanced_settings, :name => "pcm_enabled", :resource => host, :value => "false")` in your host_spec

agrare:
Just that one comment then this looks good to me

app/models/manageiq/providers/ibm_power_hmc/infra_manager/host.rb
agrare: :+1: yes looks great

spec/factories/advanced_setting.rb
vfebvre: Ok, fixed it. Interesting, I did not know factories could be defined implicitly.

agrare (approved):

Issue #43, opened 2022-02-28
Discussion : Multiple path LogicalUnit implementation


CallisteH:
From the HMC we have the possibility to have a VirtualSCSIStorage instance reachable from several paths, as suggested by the scheme defined by @dberg1.

Today in ManageIQ we try to implement VirtualSCSIStorage in the disks table. We are thinking about the best approach to inventory VirtualSCSIStorage, and more precisely, in this case, LogicalUnit (HMC object). 

![image](https://user-images.githubusercontent.com/70648532/155560198-314b7ba4-9586-4cb7-aae7-3e697c0ce886.png)

CallisteH:
Following a conversation with @Fryguy @agrare and @jaywcarman, the possibility of seeing the `volume_mapping` table was brought up.

Is there an existing provider that would do a similar job?

agrare:
So @CallisteH I think I need a little more clarity about this diagram, both LPARs and VIOSs are VMs so do some VMs connect to an MPIO through other VMs?  It looks like the two VIOs' vscsi adapters are connected to the LPAR vscsi adapter which is then connected to the MPIO

dberg1:
VIOSes are special VMs that have direct disk access and serve disk I/Os to other VMs (LPARs).
LPARs access disks through the VIOSes.

agrare:
Oh okay interesting, would it be safe to say that a VIOS disk backing is an LPAR disk?

agrare:
We could use the `GuestDevice` model which is what we use to track a VM's virtual hardware, currently they track NICs but could easily be extended to track vSCSI controllers.  I've wanted to track virtual storage controllers on vmware for a while so this seems like a good time to do that.

CallisteH:
To simplify, I would say that in the HMC there is a relationship LPAR <--OneToMany--> VIOS <--ManyToOne--> LogicalUnit (through VSCSI Mappings).
On MIQ, both VIOSes and LPARs are in the `vm` table. This `vm` table has a OneToMany relationship with `hardwares`. I think we may have a problem in the relationship between VM and Hardware (OneToOne) for the specific case of the HMC. If we consider a concrete case of multipath where we have LPAR 1 linked to two vioses (VIOS1 and VIOS2) which are linked to a single LUN logical unit, as the diagram below suggests:
**Diagram of a storage infrastructure with multipath**
[...]

PR #62, opened 2022-06-14
[WIP] Adding PhysicalVolume, VirtualDisk, VirtualOpticalMedia inventory

CallisteH:
Implementation of PhysicalVolume, VirtualDisk, VirtualOpticalMedia 
image
Image above is from HMC redbook, it shows the mapping design


CallisteH:
image
According to codeclimate, some refactoring seems possible

CallisteH:
I think there is multipath for just about every object. Redundancy is used a lot in the HMC.

To finish this pull-request, it would be necessary to resolve this discussion #43

PR #63, opened 2022-06-20
Initial VM provisioning from template work

vfebvre:
Aim: provision an LPAR from an HMC template.

First step is to be able to provision an LPAR with minimum customization (LPAR name).

### DIFF
--- a/app/models/manageiq/providers/ibm_power_hmc/infra_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/infra_manager/provision/cloning.rb
@@ -3,8 +3,13 @@ module ManageIQ::Providers::IbmPowerHmc::InfraManager::Provision::Cloning
     $ibm_power_hmc_log.info("#{self.class}##{__method__}")
     raise MiqException::MiqProvisionError, "VM capture to template failed" if phase_context[:new_vm_ems_ref].nil?
 
+    if source.template?
+      association = :vms
+    else
+      association = :miq_templates
+    end
     target = InventoryRefresh::Target.new(:manager     => source.ext_management_system,
-                                          :association => :miq_templates,
+                                          :association => association,
                                           :manager_ref => {:ems_ref => phase_context[:new_vm_ems_ref]})
     EmsRefresh.queue_refresh(target)
     true
@@ -12,20 +17,36 @@ module ManageIQ::Providers::IbmPowerHmc::InfraManager::Provision::Cloning
 
   def find_destination_in_vmdb(ems_ref)
     $ibm_power_hmc_log.info("#{self.class}##{__method__}")
-    ManageIQ::Providers::IbmPowerHmc::InfraManager::Template.find_by(:name => dest_name, :ems_ref => ems_ref)
+    if source.template?
+      ManageIQ::Providers::IbmPowerHmc::InfraManager::Lpar.find_by(:name => dest_name, :ems_ref => ems_ref)
+    else
+      ManageIQ::Providers::IbmPowerHmc::InfraManager::Template.find_by(:name => dest_name, :ems_ref => ems_ref)
+    end
   end
 
   def prepare_for_clone_task
     $ibm_power_hmc_log.info("#{self.class}##{__method__}")
-    {:name => dest_name}
+    if source.template?
+      {:name => dest_name, :target_sys_uuid => dest_host.ems_ref}
+    else
+      {:name => dest_name}
+    end
   end
 
   def log_clone_options(_clone_options)
-    $ibm_power_hmc_log.info("Provisioning [#{source.name}] to [#{dest_name}]")
+    if source.template?
+      $ibm_power_hmc_log.info("Deploying template [#{source.name}] to LPAR [#{dest_name}] on host [#{dest_host.name}]")
+    else
+      $ibm_power_hmc_log.info("Capturing LPAR [#{source.name}] to template [#{dest_name}]")
+    end
   end
 
   def start_clone(clone_options)
     $ibm_power_hmc_log.info("start_clone")
-    source.make_template(clone_options[:name])
+    if source.template?
+      source.provision_lpar(clone_options[:name])
+    else
+      source.make_template(clone_options)
+    end
   end
 end
--- a/app/models/manageiq/providers/ibm_power_hmc/infra_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/infra_manager/provision/cloning.rb
@@ -23,15 +23,25 @@ module ManageIQ::Providers::IbmPowerHmc::InfraManager::Provision::Cloning
   def prepare_for_clone_task
     $ibm_power_hmc_log.info("#{self.class}##{__method__}")
     if source.template?
-      {:name => dest_name, :target_sys_uuid => ManageIQ::Providers::IbmPowerHmc::InfraManager::Host.where(:ems_id => source.ems_id).first!.uid_ems}
+      {
+        :name                => dest_name,
+        :host_id             => get_option(:placement_host_name),
+        :vlan                => get_option(:vlan)
+      }
     else
-      {:name => dest_name}
+      {
+        :name => dest_name
+      }
     end
   end
 
-  def log_clone_options(_clone_options)
+  def log_clone_options(clone_options)
     if source.template?
-      $ibm_power_hmc_log.info("Deploying template [#{source.name}] to LPAR [#{dest_name}] on host [#{dest_host}]")
+      $ibm_power_hmc_log.info("Provisioning [#{source.name}] to [#{clone_options[:name]}]")
+      $ibm_power_hmc_log.info("Source Template:            [#{source.name}]")
+      $ibm_power_hmc_log.info("Destination VM Name:        [#{clone_options[:name]}]")
+      $ibm_power_hmc_log.info("Destination Host:           [#{clone_options[:host_id]}]")
+      $ibm_power_hmc_log.info("Destination Vlan:           [#{clone_options[:vlan]}]")
     else
       $ibm_power_hmc_log.info("Capturing LPAR [#{source.name}] to template [#{dest_name}]")
     end
--- a/app/models/manageiq/providers/ibm_power_hmc/infra_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/infra_manager/provision/cloning.rb
@@ -24,9 +24,9 @@ module ManageIQ::Providers::IbmPowerHmc::InfraManager::Provision::Cloning
     $ibm_power_hmc_log.info("#{self.class}##{__method__}")
     if source.template?
       {
-        :name                => dest_name,
-        :host_id             => get_option(:placement_host_name),
-        :vlan                => get_option(:vlan)
+        :name    => dest_name,
+        :host_id => get_option(:placement_host_name),
+        :vlan    => get_option(:vlan)
       }
     else
       {
--- a/app/models/manageiq/providers/ibm_power_hmc/infra_manager/provision/cloning.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/infra_manager/provision/cloning.rb
@@ -23,7 +23,7 @@ module ManageIQ::Providers::IbmPowerHmc::InfraManager::Provision::Cloning
   def prepare_for_clone_task
     $ibm_power_hmc_log.info("#{self.class}##{__method__}")
     if source.template?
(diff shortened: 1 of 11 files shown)


app/models/manageiq/providers/ibm_power_hmc/infra_manager/provision/cloning.rb
```diff
@@ -3,11 +3,12 @@ def clone_complete?
...
-      association = :vms
-    else
-      association = :miq_templates
-    end
+    association =
+      if source.template?
+        :vms
+      else
+        :miq_templates
+      end
```
agrare, line 11: For something this simple we usually prefer to use the ternary operator

```suggestion
    association = source.template? ? :vms : :miq_templates
```

spec/models/manageiq/providers/ibm_power_hmc/infra_manager/provision_workflow_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      expect(workflow.allowed_hosts_obj).to match_array([@host1, @host2])
+    end
+  end
+
+  context '#allowed_hosts_obj_no_stubs' do
+    let(:workflow) { described_class.new({}, admin.userid) }
+
+    before do
+      stub_dialog(:get_dialogs)
+      workflow.instance_variable_set(:@values, :src_vm_id => template.id)
```
vfebvre, line 38: Test passes (this test case comes from the VMware infra provider).

spec/models/manageiq/providers/ibm_power_hmc/infra_manager/provision_workflow_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+
+    it 'finds all hosts with no selected network' do
+      expect(workflow.allowed_hosts_obj).to match_array([host])
```
vfebvre, line 42: Test fails (returns empty array)

spec/models/manageiq/providers/ibm_power_hmc/infra_manager/provision_workflow_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+    end
+
+    it "find_all_ems_of_type" do
+      expect(workflow.find_all_ems_of_type(Host)).to match_array([host])
```
vfebvre, line 46: Test also fails

vfebvre:
@agrare I have added two test cases `spec/models/manageiq/providers/ibm_power_hmc/infra_manager/provision_workflow_spec.rb` that fail to return a list of Hosts, through respective calls to `find_all_ems_of_type` and `allowed_hosts_obj`. Not sure if this is expected or not.

Anyway on the UI form, the list of Hosts is still empty:
image

agrare:
@vfebvre I believe the issue is that you refresher isn't creating a Datacenter and Cluster then building the ems_metadata tree.

I think the move here is to override the `def allowed_hosts_obj(options = {})` method and drop all of the datacenter references and replace `hosts_ids = find_all_ems_of_type(Host).collect(&:id)` with `load_ar_obj(src[:ems]).hosts.pluck(:id)`

vfebvre:
> @vfebvre I believe the issue is that you refresher isn't creating a Datacenter and Cluster then building the ems_metadata tree.
> 
> I think the move here is to override the `def allowed_hosts_obj(options = {})` method and drop all of the datacenter references and replace `hosts_ids = find_all_ems_of_type(Host).collect(&:id)` with `load_ar_obj(src[:ems]).hosts.pluck(:id)`

@agrare This works for `allowed_hosts_obj` but it seems `allowed_hosts` ends up calling `find_all_ems_of_type` in the end to return an empty array. Is there something else I should override, or is there something missing in our inventory ?

app/models/manageiq/providers/ibm_power_hmc/infra_manager/provision/cloning.rb
```diff
@@ -3,29 +3,56 @@ def clone_complete?
...
   end
 
   def find_destination_in_vmdb(ems_ref)
     $ibm_power_hmc_log.info("#{self.class}##{__method__}")
-    ManageIQ::Providers::IbmPowerHmc::InfraManager::Template.find_by(:name => dest_name, :ems_ref => ems_ref)
+    if source.template?
+      ManageIQ::Providers::IbmPowerHmc::InfraManager::Lpar.find_by(:name => dest_name, :ems_ref => ems_ref)
+    else
+      ManageIQ::Providers::IbmPowerHmc::InfraManager::Template.find_by(:name => dest_name, :ems_ref => ems_ref)
+    end
```
agrare, line 20: I see this was like this before, but we really need to be including the ems_id here.  You can either add it to the `find_by` like `ManageIQ::Providers::IbmPowerHmc::InfraManager::Lpar.find_by(:ext_management_system => source.ext_management_system, :name => dest_name, :ems_ref => ems_ref)` or use the association like `source.ext_management_system.vms.find_by(:name => dest_name, :ems_ref => ems_ref)`

Also if the ems_ref is unique (it needs to be) then you don't need the name in the query.

app/models/manageiq/providers/ibm_power_hmc/infra_manager/provision/cloning.rb
```diff
@@ -3,29 +3,56 @@ def clone_complete?
...
 
-  def log_clone_options(_clone_options)
-    $ibm_power_hmc_log.info("Provisioning [#{source.name}] to [#{dest_name}]")
+  def log_clone_options(clone_options)
+    if source.template?
+      $ibm_power_hmc_log.info("Provisioning [#{source.name}] to [#{clone_options[:name]}]")
+      $ibm_power_hmc_log.info("Source Template:            [#{source.name}]")
+      $ibm_power_hmc_log.info("Destination VM Name:        [#{clone_options[:name]}]")
+      $ibm_power_hmc_log.info("Destination Host:           [#{clone_options[:host_id]}]")
+      $ibm_power_hmc_log.info("Destination Vlan:           [#{clone_options[:vlan]}]")
```
agrare, line 44: Should either align these to the `[#{clone_options[:name]}]` above or to the `Destination VM Name: ` here

app/models/manageiq/providers/ibm_power_hmc/infra_manager/template.rb
```diff
@@ -1,2 +1,31 @@
 class ManageIQ::Providers::IbmPowerHmc::InfraManager::Template < ManageIQ::Providers::InfraManager::Template
+  supports :provisioning
+
+  def get_host(host_id)
+    ManageIQ::Providers::IbmPowerHmc::InfraManager::Host.find(host_id)
+  end
```
agrare, line 6: This doesn't make much sense as an instance method (especially a public one) since it isn't using any instance-level attributes.  If you have an exact host_id you could just do `Host.find(options[:host_id])` in the method below

spec/models/manageiq/providers/ibm_power_hmc/infra_manager/provision_workflow_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+      EvmSpecHelper.local_miq_server
+
+      @host1  = FactoryBot.create(:ibm_power_hmc_host, :ext_management_system => ems, :ems_ref => "12345", :power_state => "on")
+      @host2  = FactoryBot.create(:ibm_power_hmc_host, :ext_management_system => ems, :ems_ref => "67890", :power_state => "on")
+      @src_vm = FactoryBot.create(:ibm_power_hmc_lpar, :ext_management_system => ems, :ems_ref => "3F3D399B-DFF3-4977-8881-C194AA47CD3A", :host => @host1)
+      stub_dialog(:get_dialogs)
+      workflow.instance_variable_set(:@values, :vm_tags => [], :src_vm_id => @src_vm.id)
+      workflow.instance_variable_set(:@target_resource, nil)
+
+      allow(workflow).to receive(:find_all_ems_of_type).and_return([@host1, @host2])
```
agrare, line 21: Are you using `find_all_ems_of_type` still?  Thought you bypassed that above

spec/models/manageiq/providers/ibm_power_hmc/infra_manager/provision_workflow_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  context '#allowed_hosts_obj_no_stubs' do
+    let(:workflow) { described_class.new({:src_vm_id => template.id}, admin.userid) }
+
+    before do
+      stub_dialog(:get_dialogs)
+    end
+
+    it 'finds all hosts with no selected network' do
+      host
```
agrare, line 41: If you want to eagerly define `host` you can use `let!(:host)` above

app/models/manageiq/providers/ibm_power_hmc/infra_manager/template.rb
```diff
@@ -1,2 +1,31 @@
...
+
+  def get_vlan(host, vlan_name)
+    Lan.find_by(:ems_ref => host.ems_ref, :name => vlan_name)
+  end
+
+  def provision_lpar(options)
+    $ibm_power_hmc_log.info("#{self.class}##{__method__} template_name #{name} lpar name #{options[:name]}")
+    host = get_host(options[:host_id])
+    vlan = get_vlan(host, options[:vlan])
```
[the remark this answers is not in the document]
vfebvre: They are

spec/models/manageiq/providers/ibm_power_hmc/infra_manager/provision_workflow_spec.rb
vfebvre: Ok, I still have to review existing specs and add some more

spec/models/manageiq/providers/ibm_power_hmc/infra_manager/provision_workflow_spec.rb
vfebvre:
```
let!(:host)     { FactoryBot.create(:ibm_power_hmc_host, :ext_management_system => ems, :ems_ref => "d47a585d-eaa8-3a54-b4dc-93346276ea37") }
...
     it 'finds all hosts with no selected network' do
      expect(workflow.allowed_hosts).to match_array([workflow.host_to_hash_struct(host)])
     end
```

spec/models/manageiq/providers/ibm_power_hmc/infra_manager/provision_workflow_spec.rb
agrare: Yeah I am able to remove this line and the specs still pass, and changing from `allow()` to `expect` shows it is called 0 times

spec/models/manageiq/providers/ibm_power_hmc/infra_manager/provision_workflow_spec.rb
vfebvre: let! apparently forces re-evaluation before each example so that explains it. With the way the specs are arranged, I think I can only define host as an instance variable in a before block.

agrare: @vfebvre actually `before` and `@host` behaves the exact same way, `before` is run before each `it`

app/models/manageiq/providers/ibm_power_hmc/infra_manager/provision/cloning.rb
```diff
@@ -14,9 +14,9 @@ def clone_complete?
   def find_destination_in_vmdb(ems_ref)
     $ibm_power_hmc_log.info("#{self.class}##{__method__}")
     if source.template?
-      ManageIQ::Providers::IbmPowerHmc::InfraManager::Lpar.find_by(:name => dest_name, :ems_ref => ems_ref)
+      source.ext_management_system.vms.find(source.id)
```
agrare, line 17: Hm don't think this is right, we are trying to find the destination by ems_ref (we don't know the dest ID yet)

app/models/manageiq/providers/ibm_power_hmc/infra_manager/provision/cloning.rb
```diff
@@ -14,9 +14,9 @@ def clone_complete?
...
     else
-      ManageIQ::Providers::IbmPowerHmc::InfraManager::Template.find_by(:name => dest_name, :ems_ref => ems_ref)
+      source.ext_management_system.miq_templates.find(source.id)
```
agrare, line 19:
```suggestion
      source.ext_management_system.miq_templates.find_by(:ems_ref => ems_ref)
```

agrare (approved):

agrare:
@vfebvre ping me when you take this out of WIP

PR #64, opened 2022-06-23
Handle case where VM or Template does not have any OS info

vfebvre:
This can happen for Templates, where OS information can be left out. This generates a nil dereferencing.


jaywcarman (approved):
Thanks @vfebvre !

PR #65, opened 2022-06-28
[WIP] Feature/mappings inventory

CallisteH:


CallisteH:
I don't know yet why it records only PhysicalVolumes.

app/models/manageiq/providers/ibm_power_hmc/inventory/parser/infra_manager.rb
```diff
@@ -270,6 +273,20 @@ def parse_templates
...
+  def parse_mappings(vios, hardware)
+    # $ibm_power_hmc_log.info("#{self.class}##{__method__} VIOS to analyze : #{vios}")
+    collector.vscsi_mappings_with_storage[vios.name].each do |mapping|
+      $ibm_power_hmc_log.info("#{self.class}##{__method__} this mapping from vios #{vios.name} will be added : #{mapping.storage}")
+      persister.guest_devices.build(
+        :name        => mapping.storage.name,
+        :size        => mapping.storage.kind_of?(IbmPowerHmc::VirtualOpticalMedia) ? mapping.storage.size : mapping.storage.capacity,
+        :uid_ems     => vios.id,
+        :device_type => mapping.storage.class.name,
+        :hardware    => hardware
```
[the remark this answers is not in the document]
CallisteH: One Hardware can have multiple guest_devices so it's probably not the right assumption.
https://github.com/CallisteH/manageiq/blob/03cb63e2c1b255600c7a48837423aa21aec87ef8/app/models/hardware.rb#L20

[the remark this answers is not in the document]
agrare: Correct a hardware is 1:1 with a vm/host but can have many guest_devices.  I think the issue is your uid_ems (which is what we're treating as unique [[ref]](https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/inventory/persister/builder/infra_manager.rb#L22)) is the `vios.id` where you're looping through many mappings per vios.  This is the reason you're only getting one.

app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
```diff
@@ -20,6 +20,8 @@ class ManageIQ::Providers::IbmPowerHmc::InfraManager < ManageIQ::Providers::Infr
   supports :provisioning
 
   has_many :hosts_advanced_settings, :through => :hosts, :source => :advanced_settings
+  has_many :miq_scsi_targets, -> { distinct }, :through => :guest_devices
```
CallisteH, line 23: Hi @agrare, do you think we have to create an empty class for MiqScsiTargets and MiqScsiLuns inside this repository ? Is it something conventional ?

agrare: Hey @CallisteH no you don't need to, those tables don't have STI so they can't be subclassed

app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
```diff
@@ -20,6 +20,8 @@ class ManageIQ::Providers::IbmPowerHmc::InfraManager < ManageIQ::Providers::Infr
...
+  has_many :miq_scsi_luns, -> { distinct }, :through => :miq_scsi_targets
```
agrare, line 24: @CallisteH I am moving these associations to core so that you don't have to define these here.  These can be removed once https://github.com/ManageIQ/manageiq/pull/21980 is merged.

CallisteH:
I have an interface issue and it seems to be caused by my code. If I put the repository of the provider on the master, I don't have this problem
image

CallisteH:
Example of what we get in the database : 
```
 id  | miq_scsi_target_id | lun | canonical_name | lun_type |              device_name              | block | block_size | capacity |           device_type            |      uid_ems
-----+--------------------+-----+----------------+----------+---------------------------------------+-------+------------+----------+----------------------------------+--------------------
  65 |                 65 |     |                |          | hdisk6                                |       |            |   122880 | IbmPowerHmc::PhysicalVolume      | 0x8200000000000000
  66 |                 66 |     |                |          | Fedora-Server-33                      |       |            |        1 | IbmPowerHmc::VirtualOpticalMedia | 0x8100000000000000
  67 |                 67 |     |                |          | hdisk8                                |       |            |    40960 | IbmPowerHmc::PhysicalVolume      | 0x8100000000000000
  68 |                 68 |     |                |          | hdisk6                                |       |            |   102400 | IbmPowerHmc::PhysicalVolume      | 0x8100000000000000

```

CallisteH:
Last tasks : 
- [ ] establish a link between the guest_device table (or maybe the miq_scsi_luns and miq_scsi_targets tables) with disk table. Not a relational link but just store an information related to identification

CallisteH:
Still this issue, mayebe caused by a wrong format or a lack of filled in attributes
image

agrare:
@CallisteH we're going to need the backtrace from that error in the logs and/or `rails s` in order to help out

app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
vfebvre: @agrare Not sure if this related, but I end up having new duplicate entries in the `miq_scsci_targets`/`miq_scsi_luns` tables every time I refresh the inventory. What do I have to do to ensure these entries remain unique ?

```
vmdb_development=# select * from miq_scsi_targets;
 id  | guest_device_id | target | iscsi_name | iscsi_alias | address |                  uid_ems
-----+-----------------+--------+------------+-------------+---------+--------------------------------------------
 530 |            1109 |        | vtscsi0    |             |         | 0862cc1c7985504a73
 531 |            1109 |        | vtopt0     |             |         | 195a07c107a921ac58
 532 |            1126 |        | vtscsi3    |             |         | 08d8e5257e750960dd
(...)
 544 |            1109 |        | vtscsi0    |             |         | 0862cc1c7985504a73
 545 |            1109 |        | vtopt0     |             |         | 195a07c107a921ac58
 546 |            1126 |        | vtscsi3    |             |         | 08d8e5257e750960dd
(...)
(28 rows)
````

Comment (agrare) [reply app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb]: @vfebvre there are two things that typically cause this, one is the collection's association is misconfigured so it is returning `[]` for the existing records, the other is you aren't using the manager_ref correctly.  Since you have the guest_device_id and uid_ems the same I'm guessing it is the association.

I'm thinking the issue could be the parent_inventory_collections here: https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/inventory/persister/builder/infra_manager.rb#L213

Try deleting that and see if it fixes your issue.  If so I'll move that down to VMware

Comment (agrare) [reply app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb]: Ohh right I keep forgetting these are for VMs not hosts

File: app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb [hunk 4]
```diff
@@ -39,6 +39,12 @@ def parse
         # have changed (e.g. RMCState, PartitionName, PartitionState etc...)
         # This may be used to perform quick property REST API calls to the HMC
         # instead of querying the full LPAR data.
+        if elems[:type].eql?("VirtualIOServer") && /Virtual.+Mapping/ =~ raw_event[:detail]
+          controller_type = "client #{/SCSI/ =~ raw_event[:detail] ? "scsi" : "vfc"} storage adapter"
```
agrare, line 43: If you just want to do a substring check it is slightly more performance to do a simple `raw_event[:detail].include?("SCSI")` than using a regex

app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
```diff
@@ -19,7 +19,9 @@ class ManageIQ::Providers::IbmPowerHmc::InfraManager < ManageIQ::Providers::Infr
   supports :metrics
   supports :provisioning
 
-  has_many :hosts_advanced_settings, :through => :hosts, :source => :advanced_settings
+  has_many :hosts_advanced_settings,           :through => :hosts, :source => :advanced_settings
+  has_many :miq_scsi_targets, -> { distinct }, :through => :guest_devices
+  has_many :miq_scsi_luns,    -> { distinct }, :through => :miq_scsi_targets
```
vfebvre, line 24: The :miq_scsi_luns association can probably be removed as it is identical to the one from the base class.

app/models/manageiq/providers/ibm_power_hmc/infra_manager/lpar.rb
```diff
@@ -79,6 +79,14 @@ def process_samples(counters, samples)
     end || {}
   end
 
+  def mapped_disks
+    ext_management_system.disks.joins("INNER JOIN miq_scsi_luns ON miq_scsi_luns.device_name=disks.device_name").joins("INNER JOIN miq_scsi_targets ON miq_scsi_targets.id=miq_scsi_luns.miq_scsi_target_id").joins("INNER JOIN guest_devices ON guest_devices.id=miq_scsi_targets.guest_device_id").where(guest_devices: {ems_ref: ems_ref})
+  end
+
+  def hardware
+    ManageIQ::Providers::IbmPowerHmc::InfraManager::VmHardware.new(super)
```
agrare, line 87: Is this trying to replace the `::Hardware` record that would normally be returned?

vfebvre: Yes, this is a wrapper around the hardware method.

In the HMC REST API, details of a volume can be found under the VIOS object providing access to the volume and not on the LPAR object the volume is ultimately attached to. We have preserved this relationship in the HMC provider by storing Disk entries under the Hardware of the VIOS Vm instead of the LPAR Vm, mainly for Targeted Refresh considerations.

In the UI, it would probably be more useful to list volumes under the LPAR Vm. Hence attempt at redirecting UI calls to `@record.hardware.disks` or other disk-related methods to pass through new method `mapped_disks` that returns list of disks actually attached to the VM.

CallisteH:
Fixed by https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/pull/91

PR #70, opened 2022-08-04
[Najdorf] Explicitly declare Template cloning as not supported

vfebvre:
The Template view is leaving Template cloning enabled even though this is not supported in this version. Clicking on this menu item leaves the user with a blank form.

image

The menu item is checking the list of supported features of the record

https://github.com/ManageIQ/manageiq-ui-classic/blob/najdorf/app/helpers/application_helper/button/generic_feature_button.rb
https://github.com/ManageIQ/manageiq-ui-classic/blob/c98fb3d7091db45a707b332ab7b7949119c2e38a/app/helpers/application_helper/toolbar/x_miq_template_center.rb#L98
https://github.com/ManageIQ/manageiq-ui-classic/blob/c98fb3d7091db45a707b332ab7b7949119c2e38a/app/helpers/application_helper.rb#L203

image

By explicitly declaring Template cloning as not supported, the menu item is disabled.

### DIFF
--- a/app/models/manageiq/providers/ibm_power_hmc/infra_manager/template.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/infra_manager/template.rb
@@ -1,2 +1,3 @@
 class ManageIQ::Providers::IbmPowerHmc::InfraManager::Template < ManageIQ::Providers::InfraManager::Template
+  supports_not :clone
 end


agrare (approved):
This is fine for a bug fix but I think going forward we should update the base MiqTemplate class to have `supports_not :clone` (https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_template/operations.rb#L5).

Issue #71, opened 2022-08-05
Template clone fails


vfebvre:
Template clone fails

Steps to reproduce
- Select a template from either the Template list or detailed view. From the Lifecycle menu, click on "Clone selected Template" or "Clone this template".
- Fill in the "template name" field on the "Clone Template" form.
- Submit form.

Results:
 - The list of request has a new entry wrongly described as "Publish" instead of "Clone".
 - It ends up in error.
 - EVM log:
```
[----] E, [2022-08-05T16:24:26.765649 #34267:add4] ERROR -- evm: Q-task_id([r8_miq_provision_8]) MIQ(ManageIQ::Providers::IbmPowerHmc::InfraManager::Provision#provision_error) [[NoMethodError]: undefined method ```make_template' for #<ManageIQ::Providers::IbmPowerHmc::InfraManager::Template:0x000001003c32c1b0>
```

Believed to be caused by https://github.com/ManageIQ/manageiq-ui-classic/pull/8377 : Template clone requests are now of type "clone_to_template" and not "clone_to_vm" anymore
On the HMC Provider, "clone_to_template" is currently treated as a request to create a template from a VM. It calls make_template, which is not implemented in the Template class

Fryguy:
@vfebvre Is this fixed via #72?

vfebvre:
@Fryguy Yes, correct, this issue was fixed by #72. I do not have the rights to add PRs to issues on this repo, but otherwise, this issue can be closed.

PR #72, opened 2022-08-09
Fix VM/Template request workflow

vfebvre:
Fix for https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/issues/71, following https://github.com/ManageIQ/manageiq-ui-classic/pull/8377

Reflect fact that Template clone requests are now of type "clone_to_template" and not "clone_to_vm" anymore


app/models/manageiq/providers/ibm_power_hmc/infra_manager/lpar.rb
```diff
@@ -34,6 +34,13 @@ def raw_rename(new_name)
     end
   end
 
+  def do_request(request_type, options)
+    case request_type
+    when 'clone_to_template'
```
agrare, line 39: If this is the only possible request type can we just call `make_template` directly?  Or are you skipping other request types?

agrare (approved):
Just one comment but we can tackle that in a follow-up

PR #73, opened 2022-08-11
Fix Psych warning regarding '-'

Fryguy:
Psych warns:

    ~/.rubies/ruby-2.7.6/lib/ruby/2.7.0/psych/visitors/to_ruby.rb:106: warning: character class has '-' without escape: /\A(?:[A-Za-z0-9-_]+)\Z/
    ~/.gem/ruby/2.7.6/gems/activesupport-6.1.6.1/lib/active_support/core_ext/marshal.rb:8: warning: character class has '-' without escape: /\A(?:[A-Za-z0-9-_]+)\Z/

The fix here is to just put the '-' at the end of the character class.
Otherwise, the character class is unsure if the '-' is to be used as a
range indicator or not. At the end it is unambiguously not a range
indicator.

### DIFF
--- a/content/miq_dialogs/miq_provision_ibm_power_hmc_dialogs_clone_to_template.yaml
+++ b/content/miq_dialogs/miq_provision_ibm_power_hmc_dialogs_clone_to_template.yaml
@@ -34,9 +34,9 @@
           :required_method:
           - :validate_vm_name
           - :validate_regex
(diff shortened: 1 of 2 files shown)


jrafanie (approved):
LGTM

agrare (approved):

PR #76, opened 2022-09-09
Add setting to adjust HMC API request timeout

jaywcarman:
When hosts have lost their connection to an HMC, the Managed System(s) APIs will take 2+ minutes to respond. The default response timeout for the rest-client gem is 60 seconds. Adding a user configurable setting to override the API request timeout allows inventory refresh to succeed, even when the responses are slow due to disconnected systems (or possible other reasons).

~The default value of 180 seconds should be enough to allow inventory refresh to pass in most cases, but still fail in a timely manner in the event of a "real" timeout.~

After further discussion, I set the default timeout value to 60 seconds. The intention is to leave existing behavior unchanged _unless_ the user changes the setting introduced here.

Depends on:
- https://github.com/IBM/ibm_power_hmc_sdk_ruby/pull/82
- [ ] Release `ibm_power_hmc` v0.12.1 with ^

### DIFF
--- a/manageiq-providers-ibm_power_hmc.gemspec
+++ b/manageiq-providers-ibm_power_hmc.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_dependency "ibm_power_hmc", "~> 0.12.0"
+  spec.add_dependency "ibm_power_hmc", "~> 0.12.2"
 
   spec.add_development_dependency "manageiq-style"
   spec.add_development_dependency "simplecov", ">= 0.21.2"
--- a/app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
@@ -143,7 +143,14 @@ class ManageIQ::Providers::IbmPowerHmc::InfraManager < ManageIQ::Providers::Infr
   def self.raw_connect(hostname, port, userid, password, validate_ssl, validate)
     require "ibm_power_hmc"
 
-    hc = IbmPowerHmc::Connection.new(:host => hostname, :port => port, :username => userid, :password => password, :validate_ssl => validate_ssl)
+    hc = IbmPowerHmc::Connection.new(
+      :host         => hostname,
+      :port         => port,
+      :username     => userid,
+      :password     => password,
+      :validate_ssl => validate_ssl,
+      :timeout      => Settings.ems.ems_ibm_power_hmc.api_request_timeout
+    )
     if validate
       # Do a logon/logoff to verify credentials
       hc.logon
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -1,5 +1,6 @@
 :ems:
   :ems_ibm_power_hmc:
+    :api_request_timeout: 60
     :blacklisted_event_names: []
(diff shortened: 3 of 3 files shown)


manageiq-providers-ibm_power_hmc.gemspec
```diff
@@ -19,7 +19,7 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^exe/}) { |f| File.basename(f) }
   spec.require_paths = ["lib"]
 
-  spec.add_dependency "ibm_power_hmc", "~> 0.12.0"
+  spec.add_dependency "ibm_power_hmc", "~> 0.14.0"
```
jaywcarman, line 22: If I cherry-pick https://github.com/IBM/ibm_power_hmc_sdk_ruby/pull/82/commits/3fc0b69ca96aaa6d1dff998cd1bceed7c6c67cab directly onto `v0.12.0` of the `ibm_power_hmc` gem, then specs pass. Applying the patch to the _main_ branch would require us to jump from `v0.12.0` -> `v0.14.0` and requires that the refresh spec VCR be recorded again.

jaywcarman: Perhaps the best path would be to release it as `v0.12.1`?

agrare: Usually a backwards-compatible interface addition would be a minor version bump (https://semver.org/), but this could easily be seen as a bug fix as well so no issues with `v0.12.1` especially if we want to backport this.

CallisteH:
Hi @jaywcarman, before merging this PR, I suggest a fix for the pcm problem we had, I tested this and it worked : 
Replace this : [code block](https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/blob/6d0d6d20c2d8ebd85799e12e0862ad0f9d563661/app/models/manageiq/providers/ibm_power_hmc/inventory/parser/infra_manager.rb#L95-L102)

by this : 
```ruby
    if collector.pcm_enabled[sys.uuid]
      persister.hosts_advanced_settings.build(
        :resource     => host,
        :name         => "pcm_enabled",
        :display_name => _("PCM-enabled"),
        :description  => _("Performance and Capacity Monitoring data collection enabled"),
        :value        => collector.pcm_enabled[sys.uuid].aggregation,
        :read_only    => true
      )
    end
```
Maybe we can add it to this PR ? 

(worked on it with @vfebvre)

agrare:
@CallisteH can you expand on what the "pcm problem" is?  If it is unrelated to the api timeout I suggest opening an issue for it and creating a standalone PR rather than combining with this one.

CallisteH:
@dberg1 Do you have details about the pcm problem ?

agrare:
Once this goes green I'm going to merge as is and we can discuss any other issues in separate PRs

agrare (approved):

agrare:
@Fryguy on backport this will require updating the Gemfile.lock.release

Fryguy:
Backported to `oparin` in commit 89a7202dfb19a9bb577488d1eb9dfd304554f86c.

```text
commit 89a7202dfb19a9bb577488d1eb9dfd304554f86c
Author: Adam Grare <adam@grare.com>
Date:   Tue Sep 20 10:19:03 2022 -0400

    Merge pull request #76 from jaywcarman/fix_request_timeouts_when_hosts_are_not_connected
    
    Add setting to adjust HMC API request timeout
    
    (cherry picked from commit 04e1911e748a9f66d90471d593970940d0de5319)
```

PR #77, opened 2022-09-20
Insert IP address used by RMC as a VM IP address

vfebvre:
The HMC does not give us visibility above the networking data link layer, so we are only able to display MAC address for each network interface. 

However, the HMC knows one of the IP address of the VM in order to communicate with it through Resource Monitoring Control. In many cases, this would be the IP address the VM is known by, so it has been suggested the provider should display this address in ManageIQ. Subnet mask is not available.

Screen captures after modification:
image
image
image

### DIFF
--- a/app/models/manageiq/providers/ibm_power_hmc/inventory/parser/infra_manager.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/inventory/parser/infra_manager.rb
@@ -138,6 +138,7 @@ class ManageIQ::Providers::IbmPowerHmc::Inventory::Parser::InfraManager < Manage
     end
 
     parse_vm_operating_system(vm, lpar)
+    parse_vm_ip_address(lpar, hardware)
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/ibm_power_hmc/inventory/parser/infra_manager.rb
```diff
@@ -136,6 +136,7 @@ def parse_lpar_common(lpar, type)
     end
 
     parse_vm_operating_system(vm, lpar)
+    parse_vm_ip_address(lpar, hardware)
```
dberg1, line 139: Maybe rename this to `parse_vm_networks` for consistency?

vfebvre: OK

agrare (approved):

Issue #78, opened 2022-09-21
OS Info not reported correctly when it contains spaces


dberg1:
The way we parse the operating system information returned by the HMC (comes from RSCT) does not work well with spaces:
```
    os_info = lpar.os&.split
    if os_info
      persister.operating_systems.build(
        :vm_or_template => vm,
        :product_name   => os_info[0],
        :version        => os_info[1],
        :build_number   => os_info[2]
      )
    end
```
Moreover, RSCT seems to have a 32 characters limitation, so the operating system name can be truncated:
```
"Linux/Red Hat Enterprise Linux "
"Linux/Hardware Management Conso"
```
instead of (`/etc/system-release`):
```
"Linux/Red Hat Enterprise Linux CoreOS release 4.11"
"Linux/Hardware Management Console (V8R8.6.0)"
```

HMC view:
![3f9b05d4-39d8-4098-8cd1-4551d347b789](https://user-images.githubusercontent.com/48122102/191497200-96e036f8-16e4-4511-9e1d-fac734930ef1.jpg)

MiQ view:
![65f08e8c-d13c-4b40-89cb-829bc8e10dee](https://user-images.githubusercontent.com/48122102/191497270-49907f63-be09-4979-8c52-ad39f772690a.jpg)
![1f10e66a-5a48-4e90-8d32-c17c03babee2](https://user-images.githubusercontent.com/48122102/191496676-29ac4dd4-33e7-4e8f-9ec6-d426061b9808.jpg)
![d6635859-877b-4c79-8dce-6dfd3476a6b3](https://user-images.githubusercontent.com/48122102/191496692-df625399-292a-45ae-9cca-88438e81b387.jpg)

dberg1:
If we use the full `lpar.os` string for `:product_name`, it looks better:

![image](https://user-images.githubusercontent.com/48122102/191502388-7bc53e13-bb32-4028-a109-3d18f8133f83.png)

dberg1:
Actually, the 32 characters limit is a HMC one (not RSCT) and has been lifted in HMC v10:

![209244f4-6a59-42eb-a7bc-ddc0a5ac5f7e](https://user-images.githubusercontent.com/48122102/191925925-9b45489f-0817-4d6b-9054-b92c6eb6c7fc.jpg)

dberg1:
See also https://github.com/ManageIQ/manageiq/pull/22119

Fryguy:
[...]

PR #82, opened 2022-09-23
improve operating system name parser

dberg1:
See https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/issues/78.


jaywcarman (approved):
LGTM

Fryguy:
Can you add some specs for this - would be nice to see specs showing the different possibilities and I think this method can be unit tested in isolation.

dberg1:
@Fryguy Added some spec to check the operating system is parsed correctly. Is this what you had in mind?

agrare (approved):

PR #86, opened 2022-09-29
[WIP] add native console for vms

dberg1:


dberg1:
@agrare Unfortunately, this does not seem to work (nothing appears in the UI on the VM page with those changes).
Maybe because of https://github.com/ManageIQ/manageiq-ui-classic/blob/d2aabad5ab5ef11cc315dcabcd277ffa347f2543/app/helpers/application_helper/button/vm_native_console.rb#L5 ?
I tried to implement `validate_native_console_support` too but it did not help either.
What I am trying to do is to add a link to the native management console for this VM.
Maybe there is some confusion between console (management console) and virtual terminal here as it seems MIQ supports both.

agrare:
Oh right, Ovirt/RHV had their own native_console feature that I think can just be renamed to match the standard remote_consoles.
I'll look at refactoring this tomorrow/next week.

agrare:
Oh no didn't see this was still WIP, this change looks good it is only code in core, ovirt, and UI-classic that has to change and this is fine to go in as is but we can revert or put in a follow-up if we need to change anything @dberg1

dberg1:
No problem, I didn't have anything to change.

PR #88, opened 2022-09-30
rework the collector

dberg1:
Complete rework of the collector.
We've been doing it wrong from the beginning.
This rework/cleanup avoids calling the API several times in targeted refresh and is required for future work.


app/models/manageiq/providers/ibm_power_hmc/inventory/collector/infra_manager.rb
```diff
@@ -1,68 +1,13 @@
...
-      do_vlans(connection)
-      do_templates(connection)
-      do_ssps(connection)
-      do_pcm_preferences(connection)
-      $ibm_power_hmc_log.info("end collection")
-    rescue IbmPowerHmc::Connection::HttpError => e
-      $ibm_power_hmc_log.error("managed systems query failed: #{e}")
-      raise
+      @connection = connection
+      yield
```
agrare, line 6: Is this yield buying us anything?  I think you could simply either:

```ruby
def initialize(manager, _target)
  super
  @connection = manager.connect
end
```

or

```ruby
def connection
  @connection ||= manager.connect
end
```

dberg1: No because we want to close the session with the HMC when we're done with the parsing (hence the with_provider_connection).

app/models/manageiq/providers/ibm_power_hmc/inventory/collector/infra_manager.rb
```diff
@@ -1,68 +1,13 @@
...
-  def initialize(manager, target)
-    super
-    @netadapters = {}
-    @sriov_elps = {}
-    @vnics = {}
-    @ssps = {}
-    @pcm_enabled = {}
-  end
-
   def collect!
```
agrare, line 2: I didn't realize you were calling `collect!` manually in the parser [[before]](https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/pull/88/files#diff-9f4a5a239e01b764383415b3f1422553029e542f8ab3e0b7c53b7c05500e31a3L4)

If you want to pre-collect inventory (typically for performance reasons) you should define a `collect` method that will be called automatically prior to parsing (https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/base_manager/refresher.rb#L118 which calls https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/inventory.rb#L32)

agrare: Most providers don't define this and lazy fetch inventory when the parser calls it though which is what you're doing here so I think you can just delete this method entirely.

app/models/manageiq/providers/ibm_power_hmc/inventory/collector/infra_manager.rb
agrare: Ah okay didn't realize you have an explicit logoff :+1:

app/models/manageiq/providers/ibm_power_hmc/inventory/collector/infra_manager.rb
dberg1: Unfortunately, we still need to call it from the parser because we want to close the HMC session at the end.

agrare (approved):

PR #91, opened 2022-10-03
Collect LPAR disks using VSCSI mappings

dberg1:
The idea here is to refresh LPARs that have disks attached to the VIOS when the VIOS is updated.
Disks coming from other VIOSes are queried from DB.
We manage multi-pathing using the location code of the disk as a list of client adapter names.


dberg1:
Example of output with a disk having multiple paths through 2 VSCSI client adapters:

image

dberg1:
"Number of Disks" view:

image

agrare (approved):

PR #92, opened 2022-10-03
Fix poll sleep interval in event catcher and make sure we can exit the poll loop

dberg1:
There is a possibility that we never exit the poll loop if no events are ever read from the HMC:

https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/blob/9470024cba887d6e31b11abe56ec632f2dabcb55/app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_catcher/stream.rb#L23-L29

This needs to be fixed by checking `@stop_polling` outside of the `next_events` loop.

Also, the HMC already waits about 10 seconds before returning if there is no event, so take that into account in the poll interval, otherwise we may end up waiting more than `10+@poll_sleep` seconds.

### DIFF
--- a/app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_catcher/stream.rb
@@ -1,5 +1,5 @@
 class ManageIQ::Providers::IbmPowerHmc::InfraManager::EventCatcher::Stream
-  def initialize(ems, options = {})
+  def initialize(ems, _options = {})
     @ems = ems
     @last_activity = nil
     @stop_polling = false
--- a/app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_catcher/stream.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_catcher/stream.rb
@@ -1,12 +1,8 @@
 class ManageIQ::Providers::IbmPowerHmc::InfraManager::EventCatcher::Stream
-  class ProviderUnreachable < ManageIQ::Providers::BaseManager::EventCatcher::Runner::TemporaryFailure
-  end
-
   def initialize(ems, options = {})
     @ems = ems
     @last_activity = nil
     @stop_polling = false
-    @poll_sleep = options[:poll_sleep] || 20.seconds
   end
 
   def start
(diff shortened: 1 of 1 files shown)


app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_catcher/stream.rb
```diff
@@ -17,19 +14,28 @@ def stop
...
+      events = connection.next_events(false)
+      elapsed = Time.now.utc.to_i - start_time
+
+      events.each(&block)
+      break if @stop_polling
+
+      # The HMC already waits a fair amount of time before returning
+      # if there were no events pending, so substract the query time
+      # from the poll sleep interval.
+      sleep(@poll_sleep - elapsed) if @poll_sleep > elapsed
```
agrare, line 29: This is fine but just curious how long does HMC tend to wait?

agrare: Reason I'm asking is if it blocks for long enough we might not need the sleep at all

vfebvre: Based on observations on several HMCs, the HMC waits for 10 seconds before it returns in case of no events (covering for Damien today)

dberg1: Yes, it is 10 seconds:

```
2.7.2 :008 > Benchmark.measure { hc.next_events(false) }
 => #<Benchmark::Tms:0x00000100092c86e8 @label="", @real=10.05330021888949, @cstime=0.0, @cutime=0.0, @stime=0.0, @utime=0.012647000000000297, @total=0.012647000000000297>
2.7.2 :009 > Benchmark.measure { hc.next_events(false) }
 => #<Benchmark::Tms:0x000001000924bb70 @label="", @real=10.039398250170052, @cstime=0.0, @cutime=0.0, @stime=0.0, @utime=0.012808999999999848, @total=0.012808999999999848>
2.7.2 :010 > Benchmark.measure { hc.next_events(false) }
 => #<Benchmark::Tms:0x00000100091cec10 @label="", @real=10.04056040220894, @cstime=0.0, @cutime=0.0, @stime=0.01261799999999999, @utime=0.0, @total=0.01261799999999999>
```

dberg1: Interestingly, the HMC supports a hidden "timeout" option that maybe we could set to @poll_sleep:

```
2.7.2 :015 > Benchmark.measure { hc.request(:get, "/rest/api/uom/Event") }.real
 => 10.065891639096662
2.7.2 :016 > Benchmark.measure { hc.request(:get, "/rest/api/uom/Event?timeout=1") }.real
 => 1.0571004541125149
2.7.2 :014 > Benchmark.measure { hc.request(:get, "/rest/api/uom/Event?timeout=2") }.real
 => 2.05112814810127
2.7.2 :013 > Benchmark.measure { hc.request(:get, "/rest/api/uom/Event?timeout=5") }.real
 => 5.064436040818691
```

agrare: :+1: yeah up to you guys but this would be greatly cleaned up/simplified if we dropped the extra sleep

dberg1: I'll add a timeout option to the SDK.

agrare (approved):
Nice looks great!

PR #93, opened 2022-10-04
Use Quick API to get Host status first during Inventory

vfebvre:
Improve initial fix for Hosts REST API time out issue:
https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/pull/76

The initial fix consisted in making the time out value configurable.

Here, to avoid time out cases with the big "ManagedSystems" API call altogether, we make a call to the "all hosts" Quick API first, so we are able to check the connection status of each Host before making individual calls to the normal REST API for each Host that has a "normal" status. For disconnected hosts or hosts that failed authentication, we would still build a DB entry, but with the limited info subset we get from the Quick API.

image


app/models/manageiq/providers/ibm_power_hmc/inventory/parser/infra_manager.rb
```diff
@@ -33,6 +34,37 @@ def parse_cecs
...
+      mtype_model, serial = sys["MTMS"].split("*")
+      host = persister.hosts.build(
+        :uid_ems             => sys["UUID"],
+        :ems_ref             => sys["UUID"],
+        :name                => sys["SystemName"],
+        :hypervisor_hostname => sys["MTMS"],
+        :ipaddress           => sys["IPAddress"],
+        :power_state         => lookup_power_state(sys["State"]),
+        :vmm_vendor          => "ibm_power_hmc",
+        :type                => ManageIQ::Providers::IbmPowerHmc::InfraManager::Host.name
```
agrare, line 48: A. This should be built automatically by the persister builder auto_sti_class method
B. If you do have to manually provider a type name you can just use the string `"ManageIQ::Providers::IbmPowerHmc::InfraManager::Host"` going to the `class.name` is just overhead

```suggestion
```

vfebvre: Ok, I have removed the line for both available and unavailable hosts.

agrare (approved):

PR #97, opened 2022-10-10
fix error path when collecting storage pools

dberg1:
Bug introduced in https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/pull/89


jaywcarman (approved):
LGTM

PR #98, opened 2022-10-10
Add physical I/O adapters to VMs

dberg1:
In PowerVM, it is possible to assign physical I/O adapters to a partition (VIOS or LPAR) using the IOMMU.
The HMC shows the physical I/O adapters attached to a partition like this:
image

With this change, we see them from MIQ like so:
image

### DIFF
--- a/app/models/manageiq/providers/ibm_power_hmc/inventory/parser/infra_manager.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/inventory/parser/infra_manager.rb
@@ -243,6 +243,7 @@ class ManageIQ::Providers::IbmPowerHmc::Inventory::Parser::InfraManager < Manage
     persister.hardwares.build(
       :vm_or_template  => vm,
       :memory_mb       => lpar.memory,
+      :cpu_type        => "ppc64",
       :cpu_total_cores => lpar.dedicated.eql?("true") ? lpar.procs.to_i : lpar.vprocs.to_i
     )
   end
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/ibm_power_hmc/inventory/parser/infra_manager.rb
```diff
@@ -335,6 +336,20 @@ def parse_vm_guest_devices(lpar, hardware)
...
     end
+
+    # Physical adapters can be assigned to VIOSes and LPARs using IOMMU.
+    lpar.io_adapters.each do |io|
+      persister.guest_devices.build(
+        :hardware        => hardware,
+        :uid_ems         => io.dr_name,
+        :device_type     => "physical_port",
+        :controller_type => "IO",
+        :device_name     => "Adapter",
```
agrare, line 347: The `device_name` is usually something unique to each device, is `io.dr_name` something that could be added here like `"Adapter #{io.dr_name}"` ?

dberg1: Yes. The reason I did this is that it does not render very well in the UI with the same information being displayed twice since we are also using the DR name as the location (it's the slot on the bus).

image

I'm ok either way.

agrare: Ah okay, I'm fine with this then

PR #99, opened 2022-10-11
use new modify_object api to rename LPARs and VIOSes

dberg1:
Renaming was only supported for LPARs (because the HMC SDK does not have a `rename_vios` method).
The HMC SDK now exports a generic `modify_object` method that allows to modify any attribute of any supported object types.

### DIFF
--- a/app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm.rb
@@ -8,7 +8,7 @@ class ManageIQ::Providers::IbmPowerHmc::InfraManager::Vm < ManageIQ::Providers::
   end
 
   supports :rename do
-    unsupported_reason_add(:rename, _("Host is not HMC-managed")) unless host.hmc_managed
+    unsupported_reason_add(:rename, _("Host is not HMC-managed")) unless host_hmc_managed
   end
 
   supports :native_console do
--- a/app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm.rb
@@ -7,6 +7,10 @@ class ManageIQ::Providers::IbmPowerHmc::InfraManager::Vm < ManageIQ::Providers::
     unsupported_reason_add(:control, _("Host is not HMC-managed")) unless host_hmc_managed
   end
 
+  supports :rename do
+    unsupported_reason_add(:rename, _("Host is not HMC-managed")) unless host.hmc_managed
+  end
+
   supports :native_console do
     reason ||= _("VM Console not supported because VM is orphaned") if orphaned?
     reason ||= _("VM Console not supported because VM is archived") if archived?
@@ -43,7 +47,15 @@ class ManageIQ::Providers::IbmPowerHmc::InfraManager::Vm < ManageIQ::Providers::
   def raw_suspend
   end
 
-  def raw_rename
+  def raw_rename(new_name)
+    ext_management_system.with_provider_connection do |connection|
+      connection.modify_object do
+        provider_object(connection).tap { |lpar| lpar.name = new_name }
+      end
+    rescue IbmPowerHmc::Connection::HttpError => e
+      $ibm_power_hmc_log.error("error renaming LPAR #{ems_ref} to #{new_name}: #{e}")
+      raise
+    end
   end
 
   # See LogicalPartitionState.Enum (/rest/api/web/schema/inc/Enumerations.xsd)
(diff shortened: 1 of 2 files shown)


app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm.rb
```diff
@@ -7,6 +7,10 @@ class ManageIQ::Providers::IbmPowerHmc::InfraManager::Vm < ManageIQ::Providers::
     unsupported_reason_add(:control, _("Host is not HMC-managed")) unless host_hmc_managed
   end
 
+  supports :rename do
+    unsupported_reason_add(:rename, _("Host is not HMC-managed")) unless host.hmc_managed
```
agrare, line 11: Above you use `host_hmc_managed` and here you use `host.hmc_managed` I think we should be consistent and probably prefer using the delegate (if host is `nil` for example this will fail and that will work)
```suggestion
    unsupported_reason_add(:rename, _("Host is not HMC-managed")) unless host_hmc_managed
```

dberg1: Yeah that makes sense, good catch!

app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm.rb
```diff
@@ -43,7 +47,15 @@ def raw_destroy
...
   end
 
-  def raw_rename
+  def raw_rename(new_name)
+    ext_management_system.with_provider_connection do |connection|
+      connection.modify_object do
+        provider_object(connection).tap { |lpar| lpar.name = new_name }
+      end
+    rescue IbmPowerHmc::Connection::HttpError => e
+      $ibm_power_hmc_log.error("error renaming LPAR #{ems_ref} to #{new_name}: #{e}")
```
jaywcarman, line 56: Minor observation: This error could come from both Lpar and Vios, so this error message might be a little confusing. Same for the `lpar` variable on line 53 above.

jaywcarman (approved):
LGTM

app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm.rb
dberg1: Yes, we are sometimes using the term LPAR as a generic term that also includes VIOSes (VIOSes being considered a special case of logical partition in this case, if that makes sense). I am not sure what generic term would fit best. Maybe just "partition"

dberg1: The HMC XML schema has a `BasePartition` type that is inherited by `VirtualIOServer` and `LogicalPartition`:
https://github.com/IBM/ibm_power_hmc_sdk_ruby/blob/e826b32fd5693169478f9a28218c3e2098f24df8/lib/ibm_power_hmc/parser.rb#L405

app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm.rb
```diff
@@ -96,4 +101,21 @@ def cpu_mhz_available?
...
+  private
+
+  def modify_attrs(attrs = {})
+    ext_management_system.with_provider_connection do |connection|
+      connection.modify_object do
+        provider_object(connection).tap do |obj|
+          attrs.each do |key, value|
+            obj.send("#{key}=", value)
+          end
+        end
```
agrare, line 114: This is fine for now but will only work for top-level attributes.  I think longer term it would be simpler to just yield the provider_object to the caller and let it set what it wants:
```suggestion
        yield provider_object(connection)
```

```
modify_attrs { |vm| vm.name = new_name }
```

dberg1: The provider object is the SDK object. The way it works in the HMC SDK is that we define `=` accessors that modify the correct elements in the XML. This is not necessarily a top-level attribute.

agrare: What if you wanted to set `vm.foo.bar=something` so a nested attribute?  There might not be any right now but that doesn't mean there won't be in the future

dberg1: The reason I implemented this generic method is that I want to reuse it for `raw_set_description` which is very similar to `raw_rename` except it will change the `description` attribute.

agrare (approved):
Looks good for now, I think we should revisit passing a hash vs yielding the object to modify in the future

app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm.rb
jaywcarman: Makes sense - thanks for the clarification!

PR #100, opened 2022-10-11
Add HMC shared processor pools to inventory as 'resource pools'

vfebvre:
Add HMC shared processor pools to inventory as 'resource pools'

HMC view of shared processor pools for a single host:
image

ManageIQ list view
image

ManageIQ single resource pool view
image

Implementation notes:
 - CPU reserve has to be an integer in ManageIQ but the HMC can allocate processor units with a granularity of 0.01 CPU
 - ems_ref is made up from both the Host UUID and the Pool UUID. We need to pass the Host UUID from the Event Target Parser to the Targeted Collector
 - Each host has 64 shared processor pools by default. For the inventory, we ignore those that still have the default settings, i.e. name as "SharedPoolXX" and maximum CPU set to 0.


vfebvre:
@miq-bot PR rebased

vfebvre:
@miq-bot remove_label unmergeable

vfebvre:
@miq-bot wip

agrare (approved):

Issue #103, opened 2022-10-13
Investigate the use of HMC extended groups (?group=) to reduce collection time


dberg1:
The HMC REST API supports extended groups for ManagedSystem, LogicalPartition and VirtualIOServer.

https://www.ibm.com/docs/en/power10/000V-HMC?topic=apis-managed-system

> Certain attribute always make calls to the underlying layer to get the data. These attributes will not be cached and populated in REST API unless user requested it.

Specifying `?group=None` saves a lot of time by avoiding these underlying layer queries:
```
2.7.2 :004 > Benchmark.measure { hc.request(:get, "/rest/api/uom/ManagedSystem") }.real
 => 2.741922002984211
2.7.2 :007 > Benchmark.measure { hc.request(:get, "/rest/api/uom/ManagedSystem?group=None") }.real
 => 0.5908163790591061
```

```
2.7.2 :015 > Benchmark.measure { hc.request(:get, "/rest/api/uom/VirtualIOServer") }.real
 => 1.5820963489823043
2.7.2 :019 > Benchmark.measure { hc.request(:get, "/rest/api/uom/VirtualIOServer?group=None") }.real
 => 0.2367262989282608
```
Unfortunately, for VIOSes, we gather mappings and storage so we may not be able to use `?group=None` here.
We need to check if the attributes we collect for managed systems are part of extended groups or not.

dberg1:
For LPARs, I believe we can use `?group=None`. The only difference is that it does not return the bootlist which we are not using anyway.
For managed systems, we want VSwitches and VLANs so we need at least `?group=SystemNetwork`.

PR #110, opened 2022-10-24
use hmc extended group attributes to reduce collection time

dberg1:
See https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/issues/103.


agrare:
@dberg1 could you add some info in the description detailing approximately how much time is saved with this in place?  Either a percentage decrease or a number for a specific inventory you have tested with.

Issue #90, opened 2022-10-03
Do not ignore errors during inventory collection


dberg1:
Up until now, we've been ignoring errors during the inventory collection, trying to collect as much things as we possibly could.
For example, when we collect the list of vms for a particular host, we just log and ignore any exception coming from the API:
https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/blob/f675e1889c40e83c96e0e6f672fe4ae8890f668a/app/models/manageiq/providers/ibm_power_hmc/inventory/collector/infra_manager.rb#L60-L68
This is a problem because in this case, all these vms will be archived.
It is probably better to reraise the exception and fail the full refresh.
The reason we've been doing this is that it is quite common to have API queries that fail for VIOSes (for instance if the `viodaemon` is down), so it was a best effort attempt to collect as many things as we could.

![image](https://user-images.githubusercontent.com/48122102/197989953-262e096c-9917-415f-a45c-2e93ed86cd74.png)

dberg1:
With v0.20.0 of the SDK, cluster APIs (clusters, ssps, tiers) are now using `ignoreError=true` by default which helps to alleviate the issues with the `vio_daemon` being down on some of the VIOSes.

dberg1:
Fixed by PR https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/pull/113

PR #113, opened 2022-10-25
fail the full refresh on errors

dberg1:
See https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/issues/90.

It might be cleaner if the SDK was raising a different exception (like ErrorNoExist) on 404 errors rather than testing `e.status != 404` in the provider.
As we are retrieving things in separate API calls, it is possible to get 404 errors if an object is deleted during the collection (which might take several minutes).
For example, we are retrieving the list of managed system UUIDs and then retrieving each managed system individually in a separate API call. The managed system might be gone between the two calls.


dberg1:
Using SDK v0.20.0 to get the new `HttpNotFound` exception and the cluster APIs permissive by default.
This solves the issue where the collect fails because the `vio_daemon` is down on some of the VIOSes.

app/models/manageiq/providers/ibm_power_hmc/inventory/collector/infra_manager.rb
```diff
@@ -11,27 +11,27 @@ def collect!
   def hmc
     @hmc ||= begin
       connection.management_console
-    rescue IbmPowerHmc::Connection::HttpError => e
+    rescue => e
       $ibm_power_hmc_log.error("management console query failed: #{e}")
```
agrare, line 15: Completely unrelated but if all we're doing here is logging on errors I wonder if we could add a Logger to the sdk gem and set it to `$ibm_power_hmc`

agrare (approved):

app/models/manageiq/providers/ibm_power_hmc/inventory/collector/infra_manager.rb
dberg1: Good idea. Maybe we can open a separate issue to do it as part of future enhancements.
These errors are also logged by EVM but the HMC provider logs have less traffic, so it's easier to spot 😊.
They also appear in UI, but they are often truncated:

image

PR #116, opened 2022-11-10
Errors in log due to some HMC Usertask events not handled properly

vfebvre:
HMC events of type "UserTask" that have their status set to something other than "Completed" are causing are not handled well by the event target parser. This generate an error in evm.log like this one:

```
[----] E, [2022-11-09T14:39:47.411859 #218489:1705c] ERROR -- evm: MIQ(MiqAeEngine.deliver) Error delivering {:event_id=>369279, :event_stream_id=>369279, :event_type=>"ADD_URI", "User::user"=>1, "EventStream::event_stream"=>369279} for object [EmsEvent.369279] with state [] to Automate: no implicit conversion of nil into Array
```

Actual error is:
```
        2: from /home/padmin/manageiq-providers-ibm_power_hmc/app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb:48:in `parse'
        1: from /home/padmin/manageiq-providers-ibm_power_hmc/app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb:48:in `concat'
TypeError (no implicit conversion of nil into Array)
```

Fix event target parser to ignore these events and add specs.

### DIFF
--- a/app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
@@ -78,6 +78,8 @@ class ManageIQ::Providers::IbmPowerHmc::InfraManager::EventTargetParser
       else
         []
       end
+    else
+      []
     end
   end
 
--- a/app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
@@ -67,17 +67,15 @@ class ManageIQ::Providers::IbmPowerHmc::InfraManager::EventTargetParser
   end
 
   def handle_usertask(usertask)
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
```diff
@@ -78,6 +78,8 @@ def handle_usertask(usertask)
       else
         []
       end
+    else
+      []
```
agrare, line 82: We tend to not like multiple nested conditionals like this if we can avoid it.

I can see a few ways to go about this:
```
return [] unless usertask["status"].eql?("Completed")

case ...
```

Or
```
result = if usertask["status"].eql?("Completed")
  case ...
  end

result || []
```

vfebvre: Fixed

agrare (approved):

PR #118, opened 2022-11-14
add vm reconfiguration for cpu, memory and network

dberg1:
Allow to modify memory and number of virtual processors of an LPAR or a VIOS (using DR if LPAR is running and RMC is active).
Allow to add/remove client network adapters to/from LPARs.
Disks reconfiguration is not supported by this PR.

image

### DIFF
--- a/app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm/reconfigure.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm/reconfigure.rb
@@ -53,19 +53,20 @@ module ManageIQ::Providers::IbmPowerHmc::InfraManager::Vm::Reconfigure
   end
 
   def build_proc_config_spec(lpar, spec, options)
+    if lpar.dedicated == "true"
+      min, max = lpar.minimum_procs, lpar.maximum_procs
+      attr = :desired_procs
+    else
+      min, max = lpar.minimum_vprocs, lpar.maximum_vprocs
+      attr = :desired_vprocs
+    end
+
     desired_procs = options[:number_of_cpus].to_i
 
-    if lpar.dedicated == "true"
-      raise MiqException::MiqVmError, "Processor count cannot be lower than #{lpar.minimum_procs}"   if desired_procs < lpar.minimum_procs.to_i
-      raise MiqException::MiqVmError, "Processor count cannot be greater than #{lpar.maximum_procs}" if desired_procs > lpar.maximum_procs.to_i
+    raise MiqException::MiqVmError, "Processor count cannot be lower than #{min}"   if desired_procs < min.to_i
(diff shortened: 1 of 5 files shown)


dberg1:
@miq-bot add_labels enhancement

app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm/reconfigure.rb
```diff
@@ -0,0 +1,112 @@
...
+  end
+
+  def max_memory_mb
+    host ? host.hardware.memory_mb : 1_024
+  end
+
+  def build_config_spec(options)
+    $ibm_power_hmc_log.debug("building spec for #{options}")
+
+    lpar = ext_management_system.with_provider_connection { |connection| provider_object(connection) }
```
agrare, line 28: Could you just use `provider_object` directly here since it will connect automatically?
```suggestion
    lpar = provider_object
```

dberg1: Unfortunately, no, because it wouldn't log off and would leave a dangling session on the HMC.

app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm/reconfigure.rb
```diff
@@ -0,0 +1,112 @@
...
+    raise MiqException::MiqVmError, "Processor count cannot be greater than #{max}" if desired_procs > max.to_i
+
+    spec[attr] = desired_procs
+  end
+
+  def build_netadap_create_config_spec(spec, options)
+    spec[:netadap_create] = options[:network_adapter_add].map do |adapt|
+      # Retrieve LAN attributes from DB.
+      switch_ids = HostSwitch.where(:host_id => host.id).pluck(:switch_id)
+      lan = Lan.find_by(:name => adapt[:network], :switch_id => switch_ids)
```
agrare, line 76: Is the network name unique across switches?

app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm/reconfigure.rb
agrare: Okay I wonder if we should (not for this PR) get rid of the raw `provider_object` method (or at least the optional `connection` with the `connection ||=`) and just use the `with_provider_object` which can release the connection as part of the block.

https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/blob/master/app/models/manageiq/providers/ibm_power_hmc/infra_manager/lpar.rb#L6-L9

app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm/reconfigure.rb
dberg1: Yes, they are unique accross all the virtual switches of a given host.

PR #120, opened 2022-11-16
Fix targeted collection for memory Resource Pools

vfebvre:
Missing implementation in TargetCollection class meant Collector implementation was called resulting in Resource Pool being removed from Inventory.


vfebvre:
@miq-bot add-labels bug

agrare (approved):

PR #124, opened 2022-11-22
Restrict supported functionalities for archived VMs & Templates

vfebvre:
Supported functionalities for VMs & Templates of type "provisionning", "clone" and "publish" are currently set without conditions. The UI therefore lets user initiate such a request on archived VMs & Templates which inevitably results in this error:

image

Proposed fix it to check at the instance level that VM or Template is not archived (i.e. it has an EMS), so the UI rejects requets.

At best, the menu item is disabled, like so:
image

At the very least, the request fails in a clean way:
image

### DIFF
--- a/app/models/manageiq/providers/ibm_power_hmc/infra_manager/lpar.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/infra_manager/lpar.rb
@@ -1,5 +1,7 @@
 class ManageIQ::Providers::IbmPowerHmc::InfraManager::Lpar < ManageIQ::Providers::IbmPowerHmc::InfraManager::Vm
-  supports :publish
+  supports :publish do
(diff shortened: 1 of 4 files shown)


agrare (approved):

Fryguy:
Backported to `oparin` in commit d120b177755644bc3d8d3e3673fcb48d8876f30b.

```text
commit d120b177755644bc3d8d3e3673fcb48d8876f30b
Author: Adam Grare <adam@grare.com>
Date:   Tue Nov 22 07:13:40 2022 -0500

    Merge pull request #124 from vfebvre/templates
    
    Restrict supported functionalities for archived VMs & Templates
    
    (cherry picked from commit 28e7181042e5d68f9f0d3bb1bae22454721dc7ea)
```

PR #127, opened 2022-11-28
implement raw_destroy for LPARs

dberg1:
image


dberg1:
@agrare I can't see the destroy item in the menu even with the `supports :terminate`. Do you know if there is anything else to do? Looks like `supports :destroy` is already set by default.

dberg1:
Waiting on https://github.com/ManageIQ/manageiq/pull/22257 and https://github.com/ManageIQ/manageiq-ui-classic/pull/8551.

agrare (approved):

PR #128, opened 2022-12-06
collect media repositories as ISO storages

dberg1:
With PowerVM, each VIOS can have a virtual media repository containing ISO images that can then be mapped to LPARs via VSCSI:

image

Depends on:
- [x] https://github.com/ManageIQ/manageiq/pull/22239


agrare (approved):
Nice!

PR #132, opened 2023-01-26
[WIP] params_for_update schema fields for Hosts

MelsHyrule:
TODO

Verify that these are the valid endpoints for this provider


agrare (approved):

agrare:
Actually now that I look at this I don't think HMC hosts support authentication, there was no verify_credentials method and they don't do scanning

MelsHyrule:
@agrare I'll close the PR then 👌

MelsHyrule:
@agrare we may need to reopen the PR bc this provider allows for edit (or remove the flag that says this host can be edited at all) 

Screen Shot 2023-03-23 at 12 37 03 PM

jaywcarman:
That right - HMC hosts don't support any kind of auth so we don't need to be able to update.

PR #145, opened 2023-09-22
Fix provisioning dialog 'src_vm_id' method

jaywcarman:
The 'allowed_templates' method is a core method to generate a list of
templates for the selection dialog. This is explicitly necessary when
creating a catalog item since the user selects the template after
entering into the request dialog (in contrast to the 'Lifecycle >
Provision VMs using this Template' which auto-selects the template).

Without this, the following exception is raised when attempting to
definite a Power HMC VM provision catalog item:

FATAL -- : Error caught: [NoMethodError] undefined method `any?' for nil:NilClass

### DIFF
--- a/content/miq_dialogs/miq_provision_ibm_power_hmc_dialogs_template.yaml
+++ b/content/miq_dialogs/miq_provision_ibm_power_hmc_dialogs_template.yaml
@@ -23,6 +23,11 @@
           :display: :hide
           :data_type: :string
         :src_vm_id:
+          :values_from:
+            :options:
+              :tag_filters: []
+
+            :method: :allowed_templates
           :description: Name
           :required: true
           :notes:


Fryguy (approved):
@agrare Please also review.  I wonder if there's a way to catch this with specs?

agrare:
Yeah we've been bitten by these content dialog files before, I'm not sure how best to test them since they are processed by the UI controllers

jaywcarman:
> I wonder if there's a way to catch this with specs?

Would love to add this, but wasn't sure how to write specs for the dialog itself.

We currently only have specs for the HMC `ProvisionWorkflow` [here](https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/blob/master/spec/models/manageiq/providers/ibm_power_hmc/infra_manager/provision_workflow_spec.rb) - but that just stubs a dialog.

Fryguy:
I'm going to merge this anyway, since we'll have to come up with a way to do specs more comprehensively.

Fryguy:
Backported to `quinteros` in commit f2aa579c1ae8f5be2ca2147380ad3c0f957de96e.

```text
commit f2aa579c1ae8f5be2ca2147380ad3c0f957de96e
Author: Jason Frey <fryguy9@gmail.com>
Date:   Tue Sep 26 09:36:30 2023 -0400

    Merge pull request #145 from jaywcarman/fix_hmc_provisioning_dialog_src_vm_id_method
    
    Fix provisioning dialog 'src_vm_id' method
    
    (cherry picked from commit 87134884fd35db4d3c88a978b774365d707ffb0e)
```

PR #147, opened 2023-10-23
Use new resource pool cpu_core_* columns

jaywcarman:
Depends on:
- https://github.com/ManageIQ/manageiq-schema/pull/708


agrare:
@jaywcarman can you add a spec test checking the new values?

PR #160, opened 2024-07-01
Remove return from supports block

kbrock:
Part of https://github.com/ManageIQ/manageiq/pull/22898 (as a followup)

Introduced by https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/pull/153

You can not have a return in a block. It causes a LongJump error Besides, it tries to return from inside the calling block - not what we want.


jrafanie (approved):

jrafanie:
Merging, codeclimate error is on their side (unable to fetch the commit).

PR #164, opened 2024-11-19
Refresh entire provider for special LPAR UUID

jaywcarman:
The HMC uses `88888888-8888-8888-8888-888888888888` to indicate an unassigned LPAR UUID. This is an error state, but we've seen it occur in a test environment so I'm adding a special case for it to trigger a full ems refresh.

### DIFF
--- a/app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
@@ -1,6 +1,8 @@
 class ManageIQ::Providers::IbmPowerHmc::InfraManager::EventTargetParser
   attr_reader :ems_event
 
+  NO_UUID_VALUE = "88888888-8888-8888-8888-888888888888" # Returned when no UUID assigned to LPAR
+
   def initialize(ems_event)
     @ems_event = ems_event
   end
@@ -39,7 +41,11 @@ class ManageIQ::Providers::IbmPowerHmc::InfraManager::EventTargetParser
         # have changed (e.g. RMCState, PartitionName, PartitionState etc...)
         # This may be used to perform quick property REST API calls to the HMC
         # instead of querying the full LPAR data.
-        new_targets << {:assoc => :vms, :ems_ref => elems[:uuid]}
+        if elems[:uuid].eql?(NO_UUID_VALUE)
+          new_targets << ems_event.ext_management_system
+        else
+          new_targets << {:assoc => :vms, :ems_ref => elems[:uuid]}
+        end
       when "VirtualSwitch", "VirtualNetwork"
         if elems.key?(:manager_uuid)
           new_targets << {:assoc => :hosts, :ems_ref => elems[:manager_uuid]}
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
```diff
@@ -39,7 +39,11 @@ def parse
         # have changed (e.g. RMCState, PartitionName, PartitionState etc...)
         # This may be used to perform quick property REST API calls to the HMC
         # instead of querying the full LPAR data.
-        new_targets << {:assoc => :vms, :ems_ref => elems[:uuid]}
+        if elems[:uuid].eql?("88888888-8888-8888-8888-888888888888")
```
agrare, line 42: Could you move this to a constant, like `INVALID_LPAR_UUID` with a comment about what it is?

agrare (approved):

agrare:
Awesome @jaywcarman !  Can we get a spec test covering this case?

app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
```diff
@@ -39,7 +41,11 @@ def parse
         # have changed (e.g. RMCState, PartitionName, PartitionState etc...)
         # This may be used to perform quick property REST API calls to the HMC
         # instead of querying the full LPAR data.
-        new_targets << {:assoc => :vms, :ems_ref => elems[:uuid]}
+        if elems[:uuid].eql?(NO_UUID_VALUE)
+          new_targets << ems_event.ext_management_system
```
agrare, line 45: NOTE this works because we have a refresh pre-processor which checks if any of the targets are an EMS, then dropping all other targets and doing a full
https://github.com/ManageIQ/manageiq/blob/master/app/models/manageiq/providers/base_manager/refresher.rb#L220

app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
```diff
@@ -39,7 +41,11 @@ def parse
         # have changed (e.g. RMCState, PartitionName, PartitionState etc...)
         # This may be used to perform quick property REST API calls to the HMC
         # instead of querying the full LPAR data.
-        new_targets << {:assoc => :vms, :ems_ref => elems[:uuid]}
+        if elems[:uuid].eql?(NO_UUID_VALUE)
```
Fryguy, line 44: No need for `eql?`, prefer a normal `==`

```suggestion
        if elems[:uuid] == NO_UUID_VALUE
```

app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
agrare: Actually I took a crack at writing a spec for this, and because the code down here https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/pull/164/files#diff-fe5f4bef291f246bd6808eda3d3c8d1d8eceb696015b491ff70d1f7784c66a65R61-R68 modifies the targets list we end up with one `InventoryRefresh::Target` that doesn't have an association or a manager_ref

Fryguy:
Backported to `radjabov` in commit d9927f7d19540a0855d80bf7985db0d66da379c5.

```text
commit d9927f7d19540a0855d80bf7985db0d66da379c5
Author: Jason Frey <fryguy9@gmail.com>
Date:   Tue Nov 19 15:52:33 2024 -0500

    Merge pull request #164 from jaywcarman/refresh_entire_ems_on_default_lpar_uuid
    
    Refresh entire provider for special LPAR UUID
    
    (cherry picked from commit 7eec25fbe3ad612eae6f4ceb5b54d450364acfe5)
```

PR #167, opened 2024-11-20
Parse the event url with Regex not split

agrare:
This isn't the greatest regular expression of all time :D but it passes the specs.

The URLs for events that we get from the specs are:
```
https://te.st:12443/rest/api/uom/ManagedSystem/e4acf909-6d0b-3c03-b75a-4d8495e5fc49/VirtualSwitch/74CC38E2-C6DD-4B03-A0C6-088F7882EF0E
https://coophmc2:12443/rest/api/uom/ManagedSystem/d47a585d-eaa8-3a54-b4dc-93346276ea37/SharedMemoryPool/557f7755-a4dc-30de-816d-387f42dd8fd3
https://te.st:12443/rest/api/uom/ManagedSystem/977848c8-3bed-360a-c9d2-ae4b7e46b5d1
https://te.st:12443/rest/api/uom/UserTask/1d43a4a6-903c-46f8-865d-a356559bb17f
https://te.st:12443/rest/api/uom/Cluster/c1e50c27-888c-3c4d-8d4a-53a3768ea250
https://coophmc2:12443/rest/api/uom/ManagedSystem/d47a585d-eaa8-3a54-b4dc-93346276ea37/SharedProcessorPool/27aae064-2855-39a0-b4e6-e3c9b5572036
https://te.st:12443/rest/api/uom/ManagedSystem/e4acf909-6d0b-3c03-b75a-4d8495e5fc49/VirtualNetwork/74CC38E2-C6DD-4B03-A0C6-088F7882EF0E
https://te.st:12443/rest/api/uom/ManagedSystem/e4acf909-6d0b-3c03-b75a-4d8495e5fc49/LogicalPartition/74CC38E2-C6DD-4B03-A0C6-088F7882EF0E
```

Critically we have to handle the optional `ManagedSystem/d47a585d-eaa8-3a54-b4dc-93346276ea37` at the start of the url path, but if that is the whole path then we need it to match on type/uuid not manager_uuid

### DIFF
--- a/app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
@@ -20,35 +20,37 @@ class ManageIQ::Providers::IbmPowerHmc::InfraManager::EventTargetParser
     case ems_event.event_type
     when "MODIFY_URI", "ADD_URI", "DELETE_URI" # Damien: INVALID_URI?
       $ibm_power_hmc_log.info("#{self.class}##{__method__} #{ems_event.event_type} #{ems_event.full_data}")
-      elems = elem(raw_event[:data])
-      case elems[:type]
+      elems = raw_event[:data]
+        .match(%r{/rest/api/uom/(?:ManagedSystem/(?<manager_uuid>[^/]+)/)?(?<type>[^/]+)/(?<uuid>[^/]+)$})
+        &.named_captures || {}
+      case elems["type"]
       when "ManagedSystem"
-        new_targets << {:assoc => :hosts, :ems_ref => elems[:uuid]}
+        new_targets << {:assoc => :hosts, :ems_ref => elems["uuid"]}
       when "LogicalPartition", "VirtualIOServer"
         # raw_event[:detail] contains information about the properties that
         # have changed (e.g. RMCState, PartitionName, PartitionState etc...)
         # This may be used to perform quick property REST API calls to the HMC
         # instead of querying the full LPAR data.
-        if elems[:uuid] == NO_UUID_VALUE
-          $ibm_power_hmc_log.info("#{self.class}##{__method__} #{elems[:type]} Missing LPAR UUID.  Escalating to full refresh for EMS: [#{ems.name}], id: [#{ems.id}].")
+        if elems["uuid"] == NO_UUID_VALUE
+          $ibm_power_hmc_log.info("#{self.class}##{__method__} #{elems["type"]} Missing LPAR UUID.  Escalating to full refresh for EMS: [#{ems.name}], id: [#{ems.id}].")
           target_collection << ems
         else
-          new_targets << {:assoc => :vms, :ems_ref => elems[:uuid]}
+          new_targets << {:assoc => :vms, :ems_ref => elems["uuid"]}
         end
       when "VirtualSwitch", "VirtualNetwork"
-        if elems.key?(:manager_uuid)
-          new_targets << {:assoc => :hosts, :ems_ref => elems[:manager_uuid]}
+        if elems["manager_uuid"]
+          new_targets << {:assoc => :hosts, :ems_ref => elems["manager_uuid"]}
         end
       when "UserTask"
         new_targets.concat(handle_usertask(raw_event[:usertask]))
       when "Cluster"
-        new_targets << {:assoc => :storages, :ems_ref => elems[:uuid]}
+        new_targets << {:assoc => :storages, :ems_ref => elems["uuid"]}
       when "SharedProcessorPool", "SharedMemoryPool"
-        new_targets << {:assoc => :resource_pools, :ems_ref => "#{elems[:manager_uuid]}_#{elems[:uuid]}"}
+        new_targets << {:assoc => :resource_pools, :ems_ref => "#{elems["manager_uuid"]}_#{elems["uuid"]}"}
       end
 
       new_targets.each do |t|
-        $ibm_power_hmc_log.info("#{self.class}##{__method__} #{elems[:type]} uuid #{t[:ems_ref]}")
+        $ibm_power_hmc_log.info("#{self.class}##{__method__} #{elems["type"]} uuid #{t[:ems_ref]}")
         target_collection.add_target(
           :association => t[:assoc],
           :manager_ref => {:ems_ref => t[:ems_ref]}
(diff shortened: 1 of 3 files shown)


app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
```diff
@@ -20,35 +20,35 @@ def parse
     case ems_event.event_type
     when "MODIFY_URI", "ADD_URI", "DELETE_URI" # Damien: INVALID_URI?
       $ibm_power_hmc_log.info("#{self.class}##{__method__} #{ems_event.event_type} #{ems_event.full_data}")
-      elems = elem(raw_event[:data])
-      case elems[:type]
+      elems = raw_event[:data].match(/^.+\/rest\/api\/uom\/(ManagedSystem\/(?<manager_uuid>.+)\/)?(?<type>.+)\/(?<uuid>.+)$/).named_captures
```
agrare, line 23: NOTE rubocop is warning about mixing captures and named captures, specifically
`(ManagedSystem\/(?<manager_uuid>.+)\/)*` but I need the entire group to be 0 or 1 of and don't actually need to capture the whole set.

app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
```diff
@@ -20,35 +20,35 @@ def parse
...
+          $ibm_power_hmc_log.info("#{self.class}##{__method__} #{elems["type"]} Missing LPAR UUID.  Escalating to full refresh for EMS: [#{ems.name}], id: [#{ems.id}].")
           target_collection << ems
         else
-          new_targets << {:assoc => :vms, :ems_ref => elems[:uuid]}
+          new_targets << {:assoc => :vms, :ems_ref => elems["uuid"]}
         end
       when "VirtualSwitch", "VirtualNetwork"
-        if elems.key?(:manager_uuid)
-          new_targets << {:assoc => :hosts, :ems_ref => elems[:manager_uuid]}
+        if elems["manager_uuid"]
```
agrare, line 39: This changed from .key? to truthy because if it isn't matched the payload will have `{"manager_uuid" => nil}`

app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
Fryguy: DO you actually need the outer group for capturing?  if not, then make it a non-capture group with `(?:ManagedSystem...)`

Fryguy, line 23: Considering all of the escaped `/`, I'm surprised rubocop didn't suggest to use %r{}, so I will :)
```suggestion
      elems = raw_event[:data].match(%r{^.+/rest/api/uom/(ManagedSystem/(?<manager_uuid>.+)/)?(?<type>.+)/(?<uuid>.+)$}).named_captures
```

agrare: > DO you actually need the outer group for capturing?

No

>  if not, then make it a non-capture group with 

Perfect, didn't know about this

Fryguy: You might also want to do `&.named_captures`, just in case it doesn't match, and then have a exception or something if elems ends up being nil.

Fryguy: ...or whatever the old code did - probably was just hashes with nils - not sure, but should be something covered by tests.

app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_target_parser.rb
```diff
@@ -20,35 +20,37 @@ def parse
     case ems_event.event_type
     when "MODIFY_URI", "ADD_URI", "DELETE_URI" # Damien: INVALID_URI?
       $ibm_power_hmc_log.info("#{self.class}##{__method__} #{ems_event.event_type} #{ems_event.full_data}")
-      elems = elem(raw_event[:data])
-      case elems[:type]
+      elems = raw_event[:data]
+        .match(%r{^.+/rest/api/uom/(?:ManagedSystem/(?<manager_uuid>.+)/)?(?<type>.+)/(?<uuid>.+)$})
```
Fryguy, line 24: Some more more regex suggestions

- Prefer a more specific matcher than `.+`, such as `[^/]+` (assuming you want to stop at the next `/` character, this is better and won't create weird backtracking)
- Remove the starting `^.+` bit because it's not useful and you don't actually need it.  As long as you have the tail anchor, it's the same result, but more performant.

Fryguy (approved):

agrare:
@Fryguy let me check if there is a spec for an invalid url first

spec/models/manageiq/providers/ibm_power_hmc/infra_manager/test_data/logical_partition_invalid_event_data.xml
```diff
@@ -0,0 +1,26 @@
...
+            <Event:Event xmlns:Event="http://www.ibm.com/xmlns/systems/power/firmware/uom/mc/2012_10/" xmlns="http://www.ibm.com/xmlns/systems/power/firmware/uom/mc/2012_10/" xmlns:ns2="http://www.w3.org/XML/1998/namespace/k2" schemaVersion="V1_6_0">
+    <Metadata>
+        <Atom>
+            <AtomID>1afb3833-8a85-369f-977a-bb8741b1a15a</AtomID>
+            <AtomCreated>1639561179310</AtomCreated>
+        </Atom>
+    </Metadata>
+    <EventType kb="ROR" kxe="false">ADD_URI</EventType>
+    <EventID kb="ROR" kxe="false">1639561179310</EventID>
+    <EventData kxe="false" kb="ROR">https://te.st:12443/rest/api/uom/ManagedSyste/e4acf909-6d0b-3c03-b75a-4d8495e5fc49/LogicalPartition/</EventData>
```
agrare, line 21: With this the regular expression returns nil to exercise that case
```
(byebug) raw_event[:data].match(%r{/rest/api/uom/(?:ManagedSystem/(?<manager_uuid>[^/]+)/)?(?<type>[^/]+)/(?<uuid>[^/]+)$})
nil
```

Issue #180, opened 2026-01-20
[Issue] VMM version and build not available in summary section for Hosts


harsha-lokesh:
The Host summary currently doesn't have any information regarding the VMM build and version.

<img width="500" height="300" alt="Image" src="https://github.com/user-attachments/assets/e9156034-1ee8-49db-95b0-4f26e3f53c2d" />

Changes are needed in `ManageIQ::Providers::IbmPowerHmc::Inventory::Parser::InfraManager`
 
Requirement (example output) :-
VMM build - 2511091800
VMM version - V11.1.1110

PR #182, opened 2026-01-20
Add the VMM build and version in UI

harsha-lokesh:
As mentioned in issue #180, this is to fix the UI VMM build and version that is not currently seen for ibm-power-hmc providers.

Before:-
image

After:- 
Screenshot 2026-01-20 at 8 18 48 PM

@miq-bot add-label PowerHMC, bug, enhancement

Fixes https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/issues/180


agrare:
Okay @harsha-lokesh looks good, just need to add spec tests then this looks good to me.  Let me know if you want any help with that.

harsha-lokesh:
Hey @agrare, I had actually made the test change however was looking on how to run the test, since it was failing due to factory setup issue. Looks like you have already done so, Thanks for updated it.
If all else is fine we can merge

Issue #181, opened 2026-01-20
Fix the native console redirection for ManageIQ


manjunbm:
Description:
The HMC native console link should redirect to new dashboard view on the HMCs with version v10.2.1030 and later

Steps to reproduce:
Connect HMC to ManageIQ, click on the details of HMC and view the native console

Describe your environment, including
   - ManageIQ Tal
   - Environment
     - For ManageIQ: Power HMC

`@miq-bot add-label PowerHMC`

PR #183, opened 2026-01-22
Fix console url to point latest dashboard for newer HMCs

manjunbm:
This PR will fix the issue of native console redirection to older dashboard on the newer HMCs.
Before:
Screenshot 2026-01-22 at 9 46 46 PM

After:
Screenshot 2026-01-22 at 9 48 38 PM

This issue will fix issue: #181 

`@miq-bot add-label bug`

### DIFF
--- a/spec/models/manageiq/providers/ibm_power_hmc/infra_manager_spec.rb
+++ b/spec/models/manageiq/providers/ibm_power_hmc/infra_manager_spec.rb
@@ -128,7 +128,7 @@ describe ManageIQ::Providers::IbmPowerHmc::InfraManager do
       it "handles greater than or equal comparisons" do
         v11 = ems.parse_hmc_version("V11R1 1110")
         v10 = ems.parse_hmc_version("V10R2 1020")
-        threshold = ems.parse_hmc_version("V10R2 1020") 
+        threshold = ems.parse_hmc_version("V10R2 1020")
         expect(v11 >= threshold).to be true
         expect(v10 >= threshold).to be true
         expect(v10 > threshold).to be false


app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
```diff
@@ -149,7 +150,92 @@ def self.raw_connect(hostname, port, userid, password, validate_ssl, validate)
...
+
+    begin
+      current_version = parse_hmc_version(hmc_version_string)
+      threshold_version = parse_hmc_version("V10R2 1020")
+      current_version >= threshold_version
+    rescue ArgumentError
+      # If version parsing fails, default to old dashboard
+      $ibm_power_hmc_log.warn("Failed to parse HMC version '#{hmc_version_string}' for dashboard comparison")
+      false
+    end
```
agrare: Instead of re-running the version comparison logic every time we want to check if we are using the new dashboard, why not just do this in verify_credentials and save the value in the `capabilities` hash?

We do this in azure to save the events capability: https://github.com/ManageIQ/manageiq-providers-azure/blob/master/app/models/manageiq/providers/azure/manager_mixin.rb#L19-L22 and then check it like this: https://github.com/ManageIQ/manageiq-providers-azure/blob/master/app/models/manageiq/providers/azure/cloud_manager.rb#L25-L27

app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
```diff
@@ -149,7 +150,92 @@ def self.raw_connect(hostname, port, userid, password, validate_ssl, validate)
...
+    
+    version_string = version_string.to_s.strip
+    
+    if version_string.match?(/^V\d+R\d+/)
+      # IBM format: extract numbers from "V<major>R<minor> <build>"
+      # e.g., "V11R1 1110" -> "11.1.1110"
+      match = version_string.match(/V(\d+)R(\d+)\s*(\d+)?/)
+      if match
+        parts = [match[1].to_i, match[2].to_i]
+        parts << match[3].to_i if match[3]
```
agrare: It might just be personal preference, but I like using named_captures for readability just array 1, 2, 3, etc... is always hard to read

```suggestion
      version_parts = version_string.match(/V(?<major>\d+)R(?<minor>\d+)\s*(?<micro>\d+)?/)&.named_captures
      if version_parts
        version = version_parts.values_at("major", "minor", "micro")
                               .compact.map(&:to_i).join(".")
        Gem::Version.new(version)
```

app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
```diff
@@ -149,7 +150,92 @@ def self.raw_connect(hostname, port, userid, password, validate_ssl, validate)
...
+    $ibm_power_hmc_log.info("#{self.class}##{__method__}: Connection established, fetching HMC version")
+    begin
+      hmc_console = connection.management_console
+      $ibm_power_hmc_log.info("#{self.class}##{__method__}: Management console obtained, retrieving version '#{hmc_console}")
+      
+      # HMC version is split across two attributes:
+      # - hmc_console.version contains release (e.g., "V11R1")
+      # - hmc_console.sp_name contains service pack/build (e.g., "1110")
+      version_part = hmc_console.version if hmc_console.respond_to?(:version)
+      sp_part = hmc_console.sp_name if hmc_console.respond_to?(:sp_name)
```
agrare: If there isn't e.g. a `sp_name` is the accessor actually missing, or is the value just `nil`?  Wondering why the `.respond_to?` is needed

app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
```diff
@@ -149,7 +150,92 @@ def self.raw_connect(hostname, port, userid, password, validate_ssl, validate)
...
+        raise ArgumentError, "Invalid IBM HMC version format: #{version_string}"
+      end
+    else
+      # Standard numeric format: "10.2.1030.0"
+      Gem::Version.new(version_string)
+    end
+  end
+
+  def fetch_and_store_hmc_version(connection)
+    $ibm_power_hmc_log.info("#{self.class}##{__method__}: Fetching HMC version for '#{hostname}', checking for connection")
```
agrare: These are fine for development/debugging but I don't think most of these log lines are that helpful.  We have a hard enough time with how verbose our logging is :)

app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
agrare: I agree but the url of the new dashboard is literally/newdashboard so all of those concerns really apply to hmc itself

app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
```diff
@@ -150,27 +151,34 @@ def self.raw_connect(hostname, port, userid, password, validate_ssl, validate)
...
     "https://#{hostname}/#{dashboard_path}/"
   end
 
-  private
-
-  def use_new_dashboard?
-    hmc_version_string = hmc_version_from_options
-    return false if hmc_version_string.blank?
-
+def update_dashboard_capability
```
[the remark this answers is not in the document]
manjunbm: My bad !! will be fixing it

Fryguy:
LGTM - I'd be happen to merge, but I think some specs should be added, particularly around the version parsing, since a future refactor could easily break that without us noticing.

agrare:
Thanks for the fix @manjunbm !!

PR #202, opened 2026-06-15
[Fix] fix for the guest restart and shutdown from ManageIQ UI

harsha-lokesh:
# Add support declarations for guest operations on IBM Power HMC VMs

## What This PR Does

This PR fixes the "Shutdown Guest action does not apply to selected items" error that occurs when attempting to perform guest operations (Shutdown Guest, Reboot Guest) on IBM Power HMC virtual machines through the ManageIQ UI.

## Problem

The IBM Power HMC provider had implemented the underlying methods for guest operations (`raw_shutdown_guest`, `raw_reboot_guest`) but was missing the required `supports` declarations. ManageIQ's UI validates feature support before executing operations by calling `records_support_feature?()`, which checks if all selected VMs support the requested feature. Without these declarations, the validation failed and prevented the operations from executing.

## Solution

Added `supports` declarations for `:shutdown_guest` and `:reboot_guest` features in the `ManageIQ::Providers::IbmPowerHmc::InfraManager::Vm` class. These declarations delegate to the existing `:control` support check, which validates that the host is HMC-managed.

### 1. Added Support Declarations
**File**: `app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm.rb`

Added:
```ruby
supports :shutdown_guest do
  unsupported_reason(:control)
end

supports :reboot_guest do
  unsupported_reason(:control)
end
```

### 2. Added Spec Tests
**File**: `spec/models/manageiq/providers/ibm_power_hmc/infra_manager/lpar_spec.rb`

Added two new test cases:
- **"supports guest operations"** - Verifies `:shutdown_guest` and `:reboot_guest` are supported when host is HMC-managed
- **"does not support guest operations"** - Verifies these operations are NOT supported when host is not HMC-managed

## Testing

- Added RSpec tests that verify support declarations work correctly
- Verified that "Shutdown Guest" and "Reboot Guest" buttons are now functional in the UI
- Confirmed that operations are properly restricted to HMC-managed hosts
- Tested that the underlying API calls (`poweroff` with appropriate parameters) are executed correctly

## API Call Flow

1. **UI** → `vm_infra_controller.rb` → VmCommon module
2. **Validation** → `ci_processing.rb` checks `supports?(:shutdown_guest)` 
3. **Core** → `vm/operations/guest.rb` calls `raw_shutdown_guest()`
4. **Provider** → `vm.rb` calls `poweroff("operation" => "osshutdown")`
5. **LPAR** → `lpar.rb` calls `connection.poweroff_lpar()`
6. **SDK** → REST API: `POST /rest/api/uom/LogicalPartition/{uuid}/do/PowerOff` with `{"operation": "osshutdown"}`

---

@miq-bot add-label bug
@miq-bot add-label enhancement

Before Fix
image

After Fix
image

### DIFF
--- a/app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm.rb
@@ -21,11 +21,19 @@ class ManageIQ::Providers::IbmPowerHmc::InfraManager::Vm < ManageIQ::Providers::
   end
 
   supports :shutdown_guest do
-    unsupported_reason(:control)
+    if !vm_powered_on?
+      _("The VM is not powered on")
+    else
+      unsupported_reason(:control)
+    end
   end
 
   supports :reboot_guest do
-    unsupported_reason(:control)
+    if !vm_powered_on?
+      _("The VM is not powered on")
+    else
+      unsupported_reason(:control)
+    end
   end
 
   supports :native_console do
--- a/app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm.rb
+++ b/app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm.rb
@@ -20,6 +20,14 @@ class ManageIQ::Providers::IbmPowerHmc::InfraManager::Vm < ManageIQ::Providers::
     _("Host is not HMC-managed") unless host_hmc_managed
   end
 
+  supports :shutdown_guest do
+    unsupported_reason(:control)
+  end
+
+  supports :reboot_guest do
+    unsupported_reason(:control)
+  end
+
   supports :native_console do
     unsupported_reason(:action)
   end
--- a/spec/models/manageiq/providers/ibm_power_hmc/infra_manager/lpar_spec.rb
+++ b/spec/models/manageiq/providers/ibm_power_hmc/infra_manager/lpar_spec.rb
@@ -45,13 +45,21 @@ describe ManageIQ::Providers::IbmPowerHmc::InfraManager::Lpar do
       expect(vm.supports?(:stop)).to be false
       expect(vm.supports?(:suspend)).to be false
     end
-    it "supports guest operations" do
+    it "supports guest operations when powered on" do
       host.advanced_settings.create!(:name => "hmc_managed", :value => "true")
       vm.raw_power_state = "running"
-      expect(vm.supports?(:shutdown_guest)).to (be true), "unsupported reason: #{vm.unsupported_reason(:shutdown_guest)}"
-      expect(vm.supports?(:reboot_guest)).to (be true), "unsupported reason: #{vm.unsupported_reason(:reboot_guest)}"
+      expect(vm.vm_powered_on?).to be_truthy
+      expect(vm.supports?(:shutdown_guest)).to be_truthy, "unsupported reason: #{vm.unsupported_reason(:shutdown_guest)}"
+      expect(vm.supports?(:reboot_guest)).to be_truthy, "unsupported reason: #{vm.unsupported_reason(:reboot_guest)}"
     end
-    it "does not support guest operations" do
+    it "does not support guest operations when powered off" do
+      host.advanced_settings.create!(:name => "hmc_managed", :value => "true")
+      vm.raw_power_state = "not activated"
+      expect(vm.vm_powered_on?).to be_falsey
+      expect(vm.supports?(:shutdown_guest)).to be_falsey
+      expect(vm.supports?(:reboot_guest)).to be_falsey
+    end
+    it "does not support guest operations when host not HMC-managed" do
       host.advanced_settings.create!(:name => "hmc_managed", :value => "false")
       vm.raw_power_state = "running"
       expect(vm.supports?(:shutdown_guest)).to be false
--- a/spec/models/manageiq/providers/ibm_power_hmc/infra_manager/lpar_spec.rb
+++ b/spec/models/manageiq/providers/ibm_power_hmc/infra_manager/lpar_spec.rb
@@ -45,6 +45,18 @@ describe ManageIQ::Providers::IbmPowerHmc::InfraManager::Lpar do
       expect(vm.supports?(:stop)).to be false
       expect(vm.supports?(:suspend)).to be false
     end
+    it "supports guest operations" do
+      host.advanced_settings.create!(:name => "hmc_managed", :value => "true")
+      vm.raw_power_state = "running"
+      expect(vm.supports?(:shutdown_guest)).to (be true), "unsupported reason: #{vm.unsupported_reason(:shutdown_guest)}"
+      expect(vm.supports?(:reboot_guest)).to (be true), "unsupported reason: #{vm.unsupported_reason(:reboot_guest)}"
+    end
+    it "does not support guest operations" do
+      host.advanced_settings.create!(:name => "hmc_managed", :value => "false")
+      vm.raw_power_state = "running"
+      expect(vm.supports?(:shutdown_guest)).to be false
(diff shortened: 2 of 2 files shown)


app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm.rb
```diff
@@ -20,6 +20,14 @@ class ManageIQ::Providers::IbmPowerHmc::InfraManager::Vm < ManageIQ::Providers::
...
   end
 
+  supports :shutdown_guest do
+    unsupported_reason(:control)
+  end
+
+  supports :reboot_guest do
+    unsupported_reason(:control)
+  end
+
```
agrare, line 38: @harsha-lokesh what happens if the VM is powered off?  Do these still show as supported?

agrare: For reference this is VMware's:
```
    supports :shutdown_guest do
      if current_state != "on"
        _("The VM is not powered on")
      elsif tools_status == 'toolsNotInstalled'
        _("The VM tools is not installed")
      else
        unsupported_reason(:control)
      end 
    end 
```

I'm not sure if you have the same guest tools requirement for this to work but the current_state definitely applies

harsha-lokesh: @agrare you are right safer to add a check, have addressed the same.

app/models/manageiq/providers/ibm_power_hmc/infra_manager/vm.rb
```diff
@@ -20,6 +20,18 @@ class ManageIQ::Providers::IbmPowerHmc::InfraManager::Vm < ManageIQ::Providers::
     _("Host is not HMC-managed") unless host_hmc_managed
   end
 
+  supports :shutdown_guest do
+    return unsupported_reason(:control) unless supports?(:control)
+
+    _("The VM is not powered on") unless vm_powered_on?
```
agrare: `return` in a proc can cause a `LocalJumpError` and should be avoided.


```suggestion
    if !vm_powered_on?
      _("The VM is not powered on")
    else
      unsupported_reason(:control)
    end
```

(NOTE the last else works because if `:control` is supported then `unsupported_reason(:control)` will be `nil`)

harsha-lokesh: @agrare Got it thank you!, have fixed it now

spec/models/manageiq/providers/ibm_power_hmc/infra_manager/lpar_spec.rb
```diff
@@ -45,6 +45,26 @@
       expect(vm.supports?(:stop)).to be false
       expect(vm.supports?(:suspend)).to be false
     end
+    it "supports guest operations when powered on" do
+      host.advanced_settings.create!(:name => "hmc_managed", :value => "true")
+      vm.raw_power_state = "running"
+      expect(vm.vm_powered_on?).to be true
+      expect(vm.supports?(:shutdown_guest)).to (be true), "unsupported reason: #{vm.unsupported_reason(:shutdown_guest)}"
+      expect(vm.supports?(:reboot_guest)).to (be true), "unsupported reason: #{vm.unsupported_reason(:reboot_guest)}"
```
agrare: The extra parenthesis aren't needed here

```suggestion
      expect(vm.supports?(:shutdown_guest)).to be_truthy, "unsupported reason: #{vm.unsupported_reason(:shutdown_guest)}"
      expect(vm.supports?(:reboot_guest)).to be_truthy, "unsupported reason: #{vm.unsupported_reason(:reboot_guest)}"
```

spec/models/manageiq/providers/ibm_power_hmc/infra_manager/lpar_spec.rb
```diff
@@ -45,6 +45,26 @@
...
+    end
+    it "does not support guest operations when powered off" do
+      host.advanced_settings.create!(:name => "hmc_managed", :value => "true")
+      vm.raw_power_state = "not activated"
+      expect(vm.vm_powered_on?).to be false
+      expect(vm.supports?(:shutdown_guest)).to be false
+      expect(vm.supports?(:reboot_guest)).to be false
```
agrare: This matches the pattern in the rest of this spec file, but in general we use `.to be_truthy` or `.to be_falsey` since that matches the way ruby handles boolean conditional comparisons.

If you really want to check that it is exactly the value `true` or `false` then it is better to use `.to eq(false)`, but we prefer truthy / falsey for rspec comparisons

spec/models/manageiq/providers/ibm_power_hmc/infra_manager/lpar_spec.rb
```diff
@@ -45,6 +45,26 @@
       expect(vm.supports?(:stop)).to be false
       expect(vm.supports?(:suspend)).to be false
     end
+    it "supports guest operations when powered on" do
+      host.advanced_settings.create!(:name => "hmc_managed", :value => "true")
+      vm.raw_power_state = "running"
```
agrare, line 50: Not for this spec, but it is good practice to combine these types of things into `context` blocks, e.g.

```
let(:vm) { FactoryBot.create(:ibm_power_hmc_lpar, :ext_management_system => ems, :ems_ref => "3F3D399B-DFF3-4977-8881-C194AA47CD3A", :host => host, :raw_power_state => raw_power_state) }
...

context "that is managed by hmc" do
  before { host.advanced_settings.create!(:name => "hmc_managed", :value => "true") }

  context "when the lpar is running" do
    let(:raw_power_state) { "running" }
  end
end

context "that is not not managed by hmc" do
end
```

spec/models/manageiq/providers/ibm_power_hmc/infra_manager/lpar_spec.rb
harsha-lokesh: Done

spec/models/manageiq/providers/ibm_power_hmc/infra_manager/lpar_spec.rb
harsha-lokesh: Done

spec/models/manageiq/providers/ibm_power_hmc/infra_manager/lpar_spec.rb
```diff
@@ -45,6 +45,26 @@
...
+      vm.raw_power_state = "not activated"
+      expect(vm.vm_powered_on?).to be_falsey
+      expect(vm.supports?(:shutdown_guest)).to be_falsey
+      expect(vm.supports?(:reboot_guest)).to be_falsey
+    end
+    it "does not support guest operations when host not HMC-managed" do
+      host.advanced_settings.create!(:name => "hmc_managed", :value => "false")
+      vm.raw_power_state = "running"
+      expect(vm.supports?(:shutdown_guest)).to be false
+      expect(vm.supports?(:reboot_guest)).to be false
```
agrare, line 66: Can you fix these two in a follow-up?

agrare: @harsha-lokesh can you fix these in a follow up 😅

harsha-lokesh: ah yes @agrare will take care of it in the next PR i have raised :)

harsha-lokesh:
Hi @agrare I have fixed this test case in the PR:-
please check and approve https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/pull/204

PR #204, opened 2026-06-22
[TEST fix] update ManageIQ#202 test case

harsha-lokesh:


app/models/manageiq/providers/ibm_power_hmc/infra_manager/vios.rb
```diff
@@ -72,6 +94,39 @@ def self.display_name(number = 1)
...
+    return_code = error_details[:return_code]
+    error_text = error_details[:exception] || error_details[:result] || error_details[:message]
+    
+    base_message = "Power off operation '#{operation}' failed"
+    base_message += " with return code #{return_code}" if return_code
+    base_message += ". Details: #{error_text}" if error_text
+    
+    base_message
+  end
+
```
agrare: Is this a duplicate of the method above?  If so it should be moved to the shared `Vm` parent class

app/models/manageiq/providers/ibm_power_hmc/infra_manager/lpar.rb
```diff
@@ -103,6 +125,39 @@ def process_samples(counters, samples)
...
+  def update_task_on_error(params, error_message)
+    # Find the most recent MiqTask for this VM's shutdown/reboot operation
+    operation = params["operation"] == "osshutdown" ? "shutdown_guest" : "stop"
+    operation = "reboot_guest" if params["restart"] == "true"
+    
+    # Search for recent task (within last 5 minutes) for this VM and operation
+    task = MiqTask.where("name LIKE ?", "%#{name}%#{operation}%")
+                  .where("created_on > ?", 5.minutes.ago)
+                  .order(:created_on => :desc)
+                  .first
```
agrare: This doesn't look like a good idea.
Check if `:miq_task_id` is one of the parameters passed in

app/models/manageiq/providers/ibm_power_hmc/infra_manager/lpar.rb
```diff
@@ -103,6 +125,39 @@ def process_samples(counters, samples)
...
+    
+    if task
+      $ibm_power_hmc_log.info("Updating MiqTask #{task.id} with error: #{error_message}")
+      task.update_status("Finished", "Error", error_message)
```
agrare: There is a `MiqTask#error` method where you can do `task.error(error_message)`

app/models/manageiq/providers/ibm_power_hmc/infra_manager/lpar.rb
```diff
@@ -40,6 +40,28 @@ def poweron(params = {})
...
+      job = e.job
+      error_details = {
+        :status => job.status,
+        :result => job.results["result"],
+        :return_code => job.results["returnCode"],
+        :exception => job.results["ExceptionText"],
+        :message => job.message
+      }
+      
+      # Log full details
```
Fryguy: Some of these comments are not necessary, please ask the AI to remove any redundant comments where the code is self-explanatory.

app/models/manageiq/providers/ibm_power_hmc/infra_manager/vios.rb
harsha-lokesh: Will move it to vm.rb

app/models/manageiq/providers/ibm_power_hmc/infra_manager/lpar.rb
harsha-lokesh: Hi @agrare 
The issue is that when operations go through check_policy_prevent automation, the original MiqTask  gets marked complete when the automation event is queued, and then raw_shutdown_guest is called WITHOUT any task context.

Hence miq_task_id won't work because there's NO miq_task_id available when raw_shutdown_guest is called through the automation callback. 
Any other approach you might have?

agrare: You're right when it goes through the policy layer it loses the context of the miq_task.  We automatically mark the task as complete when the initial `Vm#start` returns via the initial miq_task callback.

I wonder if we can pass the initial callback through so that it is the actual `raw_*` method execution which dictates the success/failure of the task, investigation needed.

harsha-lokesh: @agrare 
We can use **shutdown_guest** directly instead of **raw_shutdown_guest** however this will skip policy check which would wrong.

Another way would be current implementation with name-based MiqTask search since it Preserves policy functionality - All policies still work

agrare (approved):

agrare:
Thanks @harsha-lokesh

PR #206, opened 2026-07-23
Fix HMC Session leak issue

shashank-gowda:
1. Fix the HMC session leak issue.
2. Move HMC information retrieval to the inventory parser.
3. Work around an HMC issue where the FCS.0021 event is generated excessively. Until an HMC fix is available, filtering out this specific event during event processing.

@mig-bot add-label bug


app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_parser.rb
```diff
@@ -1,5 +1,6 @@
 module ManageIQ::Providers::IbmPowerHmc::InfraManager::EventParser
   def self.event_to_hash(event, ems_id)
+    return nil if event.detail.include?('"messageID":"FCS.0021"')
```
Fryguy: Can you add a code comment about what this is and why it's being skipped? Is this realted to the HMC session leak issue?

Alternatively, there's a section called blacklisted_event_names in the setting that can be used to skip events. Here's an example from the vmware provider: https://github.com/ManageIQ/manageiq-providers-vmware/blob/master/config/settings.yml#L4

app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
```diff
@@ -179,89 +177,7 @@ def self.raw_connect(hostname, port, userid, password, validate_ssl, validate)
...
-      legacy_dashboard = true
-    end
-    # Store in capabilities hash
-    self.capabilities = (capabilities || {}).merge(:legacy_dashboard => legacy_dashboard)
-    save! if changed?
-  end
-
-  def use_legacy_dashboard?
-    # Simply read from cached capabilities
-    !!capabilities["legacy_dashboard"]
```
Fryguy, line 206: Is the legacy dashboard no longer a thing? If so should we remove the capability from the capabilities column? (Or will refresh automatically remove the capability eventually)

Fryguy:
@shashank-gowda  For the future this was a typo (that surprisingly resolved to a user): `@Mig-Bot add-label bug` - Instead it should be `@miq-bot add-label bug`

agrare:
@shashank-gowda this seems like it should be 3 different PRs is there a reason these have to go in together?

agrare:
> Fix the HMC session leak issue.

Maybe it is obvious, but I'm not seeing it.  What part of this fixes a session leak?  Was the way we were getting the HMC version in verify_credentials causing it and moving it to the parser fixes it?

app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
agrare: Also if we're just not supporting older versions that had the legacy dashboard we can check the version in verify_credentials to ensure a user doesn't add a version that isn't supported (though ironically you're moving this out of that method)

app/models/manageiq/providers/ibm_power_hmc/infra_manager/event_parser.rb
shashank-gowda: @Fryguy For the HMC provider, we are  currently capturing only ADD_URI, MODIFY_URI, and DELETE_URI events. The event that I filtered is received as part of the MODIFY_URI event type, it should not be included there it's a bug on HMC side.

Because of this, we cannot simply blacklist MODIFY_URI, as it has all resource modification events that we depend on. Filtering the entire MODIFY_URI event type would result in losing resource modification information (Hosts, VMs etc), so as a workaround I'm filtering out the specific unwanted event (FCS.0021) until the HMC-side issue is fixed.

app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
shashank-gowda: @Fryguy @agrare Yes, the HMC that was using the legacy dashboard URL is out of support, so we can safely ignore it. And yes, refresh automatically remove the capability.

shashank-gowda:
@agrare 
> > Fix the HMC session leak issue.
> 
> Maybe it is obvious, but I'm not seeing it. What part of this fixes a session leak? Was the way we were getting the HMC version in verify_credentials causing it and moving it to the parser fixes it?

Previously, verify_credentials validated the provider credentials by calling connect(validate: true).

The connect(validate: true) flow does the following steps:

1. Creates a connection object.
2. Logged in to the HMC and gets an api_session_token, which was stored in the connection object.
3. Logged out, the api_session_token attribute in the connection object is set to nil.

After credential validation, the same connection object was passed to fetch_and_store_hmc_version().

Problem:

When fetch_and_store_hmc_version() invoked the Management Console API, the request() [(https://github.com/IBM/ibm_power_hmc_sdk_ruby/blob/main/lib/ibm_power_hmc/apis/connection.rb#L142)](https://github.com/IBM/ibm_power_hmc_sdk_ruby/blob/main/lib/ibm_power_hmc/apis/connection.rb#L142) whether the connection object has a valid api_session_token:

Since the api_session_token had already been cleared during the earlier logout, the request method created a new HMC session. But, this newly created session was never logged off, resulting in stale sessions accumulating on the HMC.

As verify_credentials was executed repeatedly led to a session leak on the HMC.

Fix

To avoid this session leak, fetching the HMC details has been moved to the inventory parser. This avoids creating an additional session during credential validation and ensures that the session lifecycle is handled correctly.


Additionally, the event filtering change was included in this PR since it was a small one line modification.

agrare:
@shashank-gowda okay so a more targeted fix of the session leak issue would be to use `with_provider_connection` in a block and use that connection for fetch_and_store_hmc_version so the logout is after that API call.

I think it would be better to have that as its own dedicated PR because we are going to want to backport that, where dropping the legacy console we probably don't want to backport.

app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
```diff
@@ -132,8 +132,8 @@ def self.verify_credentials(args)
 
   def verify_credentials(_auth_type = nil, _options = {})
     begin
-      connection = connect(:validate => true)
-      fetch_and_store_hmc_version(connection)
+      connect(:validate => true)
+      fetch_and_store_hmc_version
```
agrare: This will connect twice right?  Would this be more efficient

```suggestion
      with_provider_connection do |connection|
        connection.logon
        fetch_and_store_hmc_version(connection)
      end
```
And then let the `ensure disconnect(connection)` from the core `with_provider_connection` handle logout

app/models/manageiq/providers/ibm_power_hmc/infra_manager.rb
Fryguy: I feel like dropping legacy dashboard URL should be a PR on it's own and not lumped into session leaking?  We might also need a database migration to update the capabilities.

agrare:
https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/pull/206/changes/8978b3a85bf15c6fdf5d29a10869014617d37ed2

Please rebase rather than merge changes from master into your PR branch.  See https://www.manageiq.org/docs/guides/developer_setup/git_workflow specifically `git config --global --bool pull.rebase true`

app/models/manageiq/providers/ibm_power_hmc/infra_manager/utility/xml_to_json_transformer.rb
```diff
@@ -0,0 +1,114 @@
+require "rexml/document"
```
Fryguy: Please avoid using Rexml - in the project we use Nokogiri instead.

agrare: I don't think this is meant to be in this branch, he's going to have to rebase but I don't want this to get lost so it should apply to https://github.com/ManageIQ/manageiq-providers-ibm_power_hmc/pull/210 as well

agrare:
@shashank-gowda you've got some git troubles here, if `git pull upstream master` isn't fixing this, one other thing you can do is to use `git rebase --onto` to "reset" the branch history before a specific commit.

e.g. `git rebase --onto=upstream/master e5ffe98bb6bb70acd6a381c672b4b6108917bbda^`

